PR #12, opened 2025-09-03
Fix Agent Web Task Follow-up

matt-greathouse:
### The Issue
The agent loop closes the budget cycle on sleep because it believes there are no outstanding child tasks. The tracking for "outstanding children" is not actually updated when a web task is spawned, so it always reads zero and closes the cycle prematurely.

### The Fix
The agent loop uses AgentBudgetManager.get_branch_depth(...) to decide whether to close the cycle when sleeping. Before, nothing incremented this, so it looked like "0 children" and closed the budget prematurely. Now, the parent branch counter increments on spawn and decrements on completion; the loop will leave the cycle active when > 0, and allow follow-up processing after the last child completes.

### DIFF
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -346,27 +346,31 @@ def process_agent_events(
                 agent_id=str(persistent_agent_id), budget_id=budget_id, depth=depth
             )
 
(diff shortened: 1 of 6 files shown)


ai-christianson:
💪 great find

ai-christianson (approved):
💪

PR #78, opened 2025-09-26
Analytics for org based actions

willbonde:


api/views.py
```diff
@@ -39,6 +40,37 @@
 logger = logging.getLogger(__name__)
 tracer = trace.get_tracer('gobii.utils')
 
+
+def _derive_task_organization(task: BrowserUseAgentTask):
```
matt-greathouse, line 44: I have another branch that adds an organization column to the BU task table so orgs can make API requests with their key. So this can be updated after that.

matt-greathouse (commented):
Do we need stuff in signals.py for creating analytics events when billing is updated? i saw you had constants for those events so wasn't sure if that was in the works

PR #92, opened 2025-09-30
Fix task credit grant by id in admin

willbonde:


willbonde:
@matt-greathouse this change only affects admin page, no end user functionality tweaked

matt-greathouse (approved):

PR #99, opened 2025-10-01
MailGun Support for Incoming

willbonde:


api/agent/comms/adapters.py
```diff
@@ -247,3 +247,125 @@ def _within_size(a: Any) -> bool:
...
+            if html_body
+            else "None"
+        )
+
+        body = working_text
+
+        if EMAIL_STRIP_REPLIES is True:
+            span.set_attribute("mailgun.strip_replies", "True")
+            attachments_meta = []
+            for att in attachments:
```
matt-greathouse, line 315: make sure attachments respect settings.MAX_FILE_SIZE

matt-greathouse (changes requested):

api/agent/comms/adapters.py
willbonde: added

PR #166, opened 2025-10-29
Agent Short Description

willbonde:
Agent Short Description for quick recognition of what an agent is doing in the dashboard


matt-greathouse (commented):
Should we remove the short description columns entirely? and replace them in favor of mini descriptions?

willbonde:
> Should we remove the short description columns entirely? and replace them in favor of mini descriptions?

Open to it - do you think we should leave and come back? Or just chop it now?

PR #879, opened 2026-04-19
Revert "Additional Signup Protection"

willbonde:
Reverts gobii-ai/gobii-platform#877


willbonde:
Having deployment issues, backing out my merge as an abundance of caution (I don't think it's related but would rather not guess)

PR #1023, opened 2026-05-20
Updating library for SEO and Performance

willbonde:
- SEO issues
- Delay for users


matt-greathouse (changes requested):
LibraryScreen is no longer lazy (main.tsx:9, was line 20)
```import { LibraryScreen } from './screens/LibraryScreen'   // was: const LibraryScreen = lazy(...)```
This pulls LibraryScreen (~400 LOC + its imports) into the main bundle for every React-mounted page — chat, audit, billing, evals, etc. The SSR fallback already covers the "no flash of loading spinner" goal for crawlers and the first paint. With lazy() retained, library users would still see the SSR fallback instantly and the React chunk would stream in behind it.
If the transition flash from SSR → React is the real concern, a <link rel="modulepreload"> hint on the library page would get you the same UX without taxing every other page's bundle.

matt-greathouse (approved):

PR #1160, opened 2026-06-15
Organization Agent Templates

willbonde:


console/organization_api_views.py
```diff
@@ -155,6 +163,82 @@ def _serialize_organization(org: Organization, membership: OrganizationMembershi
     }
 
 
+def _serialize_organization_template(template: PersistentAgentTemplate) -> dict:
+    source_agent = getattr(template, "source_agent", None)
+    created_by = getattr(template, "created_by", None)
+    return {
```
matt-greathouse, line 169: Consider returning only the fields currently used by the organization screen. The frontend appears to consume `id`, `name`, `tagline`, `category`, `sourceAgentName`, `createdBy`, and `scheduleDescription`; fields like `code`, `description`, `sourceAgentId`, `baseSchedule`, `defaultTools`, and `updatedAt` expand the API surface without a current consumer.

frontend/src/screens/OrganizationScreen.tsx
```diff
@@ -229,13 +241,87 @@ function AddMemberModal({
...
+}
+
 export function OrganizationScreen() {
   const queryClient = useQueryClient()
   const queryKey = useMemo(() => ['current-organization'] as const, [])
   const { data, error, isLoading } = useQuery({
     queryKey,
     queryFn: ({ signal }) => fetchCurrentOrganization(signal),
   })
+  const templateQueryKey = useMemo(() => ['current-organization-templates'] as const, [])
```
matt-greathouse, line 315: Could the template list data ride on the existing `current-organization` payload instead of using a second GET query? Keeping POST/DELETE/launch as separate template actions still makes sense, but folding the read model into the main org settings payload would avoid a second query key, loading/error path, and duplicate cache update path for the same screen.

matt-greathouse (commented):
A few suggestions from reading through the branch. No blocking changes from me.

console/organization_api_views.py
```diff
@@ -255,6 +339,126 @@ def patch(self, request):
...
+        try:
+            org, membership, error = _require_current_org(request)
+        except PermissionDenied as exc:
+            return _json_error(str(exc), status=404)
+        if error:
+            return error
+        if membership.role not in MEMBER_MANAGE_ROLES:
+            return _json_error("You do not have permission to manage templates.", status=403)
+
+        payload, error = _parse_json_body(request)
```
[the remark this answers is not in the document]
willbonde: This one is already handled: organization_api_views.py defines a local _parse_json_body() wrapper around _parse_json_body_or_raise() that returns (payload, error) for these JSON API views. Calling _parse_json_body_or_raise() directly here would break the tuple unpacking and bypass the JSON error response path.

PR #1346, opened 2026-07-22
Fix agent charter updates and timeline synchronization

matt-greathouse:
## Summary
- Preserve and repair agent charter newlines across configuration and prompt processing
- Improve SQLite agent configuration and batch update handling
- Synchronize agent chat timeline, status, scheduling, and tool details
- Add coverage for charter repair, event processing, SQLite updates, and frontend state behavior

## Testing
- Not run (not requested)


matt-greathouse:
## Code Review — Fix agent charter updates and timeline synchronization

Reviewed the full diff (30 files, +958/−274). Overall this is a solid, well-tested change. The central design shift — the frontend now trusts the backend's `agent_config_update` confirmation and captured `charterText`/`scheduleValue` snapshots instead of re-parsing SQL to *guess* what changed — is a real correctness improvement and removes a large, fragile parser (`agentConfigSql.ts` −154). Extracting the config-statement regexes into `sqlite_config_statements.py` cleanly breaks an import cycle risk between `event_processing`, `stop_policy`, and the sqlite tools. Test coverage is good (repair command, event processing, sqlite batch/config, and frontend registry/cache/timeline state).

No blocking issues found. A few low-severity observations and questions below.

### 🟡 Low

- **Historical timeline entries will stop rendering as config updates.** `buildAgentConfigEntry` now returns `null` unless the tool result carries a backend `agent_config_update` confirmation *and* `status === 'complete'`, and schedule/charter display now depends on the new `scheduleValue`/`charterText` metadata. Tool calls persisted **before** this PR lack those fields, so past charter/schedule updates in the timeline will fall back to a generic "Database query" entry rather than "Assignment updated"/"Schedule updated". If preserving the appearance of historical entries matters, this is worth calling out; otherwise it's an acceptable one-time display regression. (`frontend/src/components/agentChat/tooling/sqliteEntries.ts`)

- **`refreshLoadedTimelineVariantsInCache` refetches with `type: 'all'`.** `refetchQueries({ queryKey: ['agent-timeline', agentId], type: 'all' })` fires on every processing active→inactive transition and refetches *all* cached timeline variants for the agent, including inactive/unobserved ones (developer/standard/staff-context permutations). Functionally correct given the new dependence on the persisted result, but on chatty agents this could add avoidable refetches. Consider whether `type: 'active'` would suffice. (`frontend/src/hooks/useTimelineCacheInjector.ts`, `frontend/src/store/chatSlice.ts`)

- **`SELECT patch_text(...) FROM __agent_config` is now rejected as an error.** `_non_persisting_agent_config_patch` blocks any statement combining `patch_text(` + `__agent_config` that doesn't assign the charter column, returning a `config_patch_not_persisted` error. This is intentional (steering the model toward the `UPDATE ... SET charter=patch_text(...)` form), but it also blocks a legitimate read-only preview of a patched value. Given the guardrail is `retryable: true` and message-guided, this is probably the right trade-off — flagging so it's a conscious one. (`api/agent/tools/sqlite_batch.py`)

### 🔵 Info / Nits

- **Charter literal-`\n` guard is a strict count increase.** `apply_sqlite_agent_config_updates` rejects a charter change when `count_literal_newlines(current) > count_literal_newlines(baseline)`. This correctly allows edits that keep or reduce literal `\n` (including the repair path), but it would also reject a legitimate charter that intentionally documents an escaped `\n` in prose/code. Very unlikely in practice; noting for completeness. (`api/agent/tools/sqlite_agent_config.py`)

- **Repair command `ambiguous` counter has two meanings.** In `repair_agent_charter_newlines`, `ambiguous` is incremented for *every* agent that has any remaining literal `\n` after repair (including successfully-`REPAIRED` agents with leftover ambiguous newlines), while the printed `AMBIGUOUS` line is only emitted for fully-unrepairable agents (`structural_count == 0`). The summary's `ambiguous=N` and the number of `AMBIGUOUS` log lines can therefore differ, which may confuse an operator reading the output. Consider distinct labels (e.g. "unrepairable" vs "has-residual-literal-newlines"). (`api/management/commands/repair_agent_charter_newlines.py`)

- **Nice ordering fix:** moving `_annotate_agent_config_update_result` *before* the `direct_correction_reply_pending` computation, and gating that flag on `_tool_result_confirms_charter_patch` (which checks the annotated `agent_config_update` rather than a bare `status=ok`), correctly closes the "generic ok ≠ persisted" gap — and the matching system-prompt note in `prompt_context.py` reinforces it. 👍

### Testing
Per the PR description, tests were not run. The added unit tests look substantive and targeted; recommend running the `batch_sqlite`-tagged suite plus the frontend vitest suite before merge to confirm green, since the change touches shared parsing/serialization paths.


---

PR #1347, opened 2026-07-22
Add webhook management tools and agent skill

matt-greathouse:
## Summary

- Add webhook management and sending tools across agent runtime, services, and console settings
- Introduce default webhook system-skill configuration and tool metadata
- Add webhook eval scenarios and unit coverage
- Improve event and prompt context handling for webhook workflows

## Testing

- Added and updated focused unit tests and webhook eval scenarios
- Not run (not requested)

### DIFF
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -133,7 +133,6 @@ from ..tools.tool_manager import ToolCatalogEntry, execute_enabled_tool, auto_en
 from ...services.tool_blacklist import is_tool_blacklisted_for_agent, tool_blacklist_error
 from ..tools.web_chat_sender import execute_send_chat_message, _looks_like_routine_progress_message
 from ..tools.peer_dm import execute_send_agent_message
-from ..tools.webhook_sender import execute_send_webhook_event
 from ..tools.agent_variables import clear_variables, get_all_variables, replace_all_variables, substitute_variables
 from ..tools.file_export_helpers import resolve_export_target
 from ..files.filespace_service import _normalize_write_path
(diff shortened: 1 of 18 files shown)


matt-greathouse (commented):
## Automated code review

Reviewed the diff against `main` (15 files, +1291/-261). Overall this is solid, well-tested work: the webhook management tools follow the existing system-skill pattern closely, agent-scoping on lookups is correctly enforced and covered by tests (including cross-agent isolation), the `is_active`/`name`/`url` update handling correctly distinguishes "not provided" from "explicit falsy value" (`"key" in params` rather than truthy checks), and the console refactor to share `AgentWebhookService` removes a good chunk of duplicated logic without changing user-facing behavior. No blocking issues found.

### Low severity / hardening suggestions

**1. Webhook-URL redaction skips short secrets (`api/agent/tools/webhook_sender.py`, `_redact_webhook_url`)**
Path, query-value, username, and password candidates are only redacted when `len(...) >= 4`. A short token/path segment (1–3 chars) in the destination URL would survive into `last_error_message` and analytics `error_message` properties. Probably fine in practice, but worth a comment noting the threshold is a deliberate tradeoff (to avoid mass-redacting trivial substrings) rather than an oversight.

**2. Bare hostname isn't a redaction candidate (`api/agent/tools/webhook_sender.py`, `_redact_webhook_url`)**
The candidate set is built from the full URL and `path+query` (plus query values / userinfo), but never the bare host (`parsed.hostname`/`netloc`). `requests`/`urllib3` connection errors commonly surface the host on its own, e.g. `HTTPSConnectionPool(host='internal.example.com', port=443): ... Failed to resolve 'internal.example.com'`. Since this PR's whole point is to keep webhook destinations out of logs/analytics/stored error messages, that hostname would still leak in exactly the failure mode (DNS/connection errors) where `_redact_webhook_url` is most likely to run. Consider adding `parsed.hostname` (guarded by the same length check) to `candidates`.

### Info

**3. Testing**: PR description notes tests were "not run (not requested)" locally, only added/updated. Given the diff touches shared tool-routing modules used by every agent turn (`event_processing.py`, `tool_runtime.py`, `static_tools.py`, `tool_manager.py`), it'd be worth confirming CI is green (or running the affected suites) before merge.

No other correctness, security, or style issues found.


---

PR #1354, opened 2026-07-22
Improve agent state updates across chat and timeline

matt-greathouse:
## Summary
- Synchronize agent state updates through the chat WebSocket protocol and consumer.
- Keep chat, timeline, and client state caches consistent with ordered updates and metadata bridging.
- Add coverage for socket routing, protocol handling, signals, cache injection, and Agent Chat behavior.

## Testing
- Added and updated frontend and backend unit tests.
- Not run (not requested).

### DIFF
--- a/console/agent_chat/consumers.py
+++ b/console/agent_chat/consumers.py
@@ -265,6 +265,7 @@ class AgentChatSessionConsumer(AsyncJsonWebsocketConsumer):
         context_override=None,
         staff_context_override=None,
     ) -> None:
+        was_subscribed = agent_id in self.subscriptions
         try:
             agent = await self._resolve_agent(
                 self.user,
@@ -327,7 +328,8 @@ class AgentChatSessionConsumer(AsyncJsonWebsocketConsumer):
                 agent_id,
                 exc,
             )
-            await self._remove_subscription(agent_id)
+            if not was_subscribed:
+                await self._remove_subscription(agent_id)
             await self.send_json(
                 {
                     "type": "subscription.error",
--- a/console/agent_chat/consumers.py
+++ b/console/agent_chat/consumers.py
@@ -4,9 +4,12 @@ from dataclasses import dataclass
 from channels.db import database_sync_to_async
 from channels.generic.websocket import AsyncJsonWebsocketConsumer
 from django.core.exceptions import PermissionDenied
+from django.db import DatabaseError
 
 from console.agent_chat.access import resolve_agent, resolve_staff_agent
+from console.agent_chat.pending_actions import expire_pending_action_requests, list_pending_action_requests
 from console.agent_chat.realtime import user_profile_group_name, user_stream_group_name
+from console.agent_chat.timeline import build_processing_snapshot, serialize_processing_snapshot
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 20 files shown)


console/agent_chat/consumers.py
```diff
@@ -315,6 +319,33 @@ async def _subscribe(
...
 
+        try:
+            snapshot = await self._build_subscription_snapshot(agent)
+        except DatabaseError as exc:
+            logger.exception(
+                "AgentChatSessionConsumer failed to build subscription snapshot (agent=%s): %s",
+                agent_id,
+                exc,
+            )
+            await self._remove_subscription(agent_id)
```
matt-greathouse: **Correctness/resilience:** `_build_subscription_snapshot` now runs on *every* `_subscribe()` call, including a pure mode toggle (`background` → `active`) for an agent that's already in `self.subscriptions`. Previously, once an agent's groups were joined, nothing in `_subscribe` could tear that membership down again.

Now, if this snapshot rebuild hits a transient `DatabaseError` (e.g. a brief DB blip) while refreshing an *already-established* subscription, `_remove_subscription(agent_id)` is called, which `group_discard`s the channel from groups it was previously — and successfully — subscribed to. The client only removes the agent from `requestedSubscriptionsRef` on `subscription_error` (see `useAgentChatSocket.ts`), and won't automatically retry unless `desiredSubscriptions` changes or the socket reconnects. So a one-off DB error on a mode-change refresh can silently and durably kill realtime delivery for that agent, which wasn't previously possible.

Consider only tearing down groups when the *initial* join (permission check / group_add) fails, and treating a snapshot-rebuild failure on an already-subscribed agent as a soft error (log + send `subscription.error` without unsubscribing), or retrying the snapshot rather than dropping the subscription.

---

console/agent_chat/consumers.py
```diff
@@ -315,6 +319,33 @@ async def _subscribe(
         elif self.active_agent_id == agent_id:
             self.active_agent_id = None
 
+        try:
+            snapshot = await self._build_subscription_snapshot(agent)
```
matt-greathouse, line 324: **Performance:** `_build_subscription_snapshot` (3 bulk `UPDATE`s via `expire_pending_action_requests`, plus a fresh `build_processing_snapshot` + `list_pending_action_requests` query set) now runs on every `subscribe` message, not just the first one for a given agent — including pure `active`/`background` mode toggles on an agent that's already subscribed (e.g. every time the user switches which chat panel is focused in a multi-agent roster).

That's a meaningful amount of extra DB work per UI interaction that previously was a no-op once subscribed. Worth confirming this load is acceptable at expected roster sizes/switch frequency, or consider skipping the snapshot rebuild when `agent_id` is already subscribed and only the mode changed (the `mode.ready` payload could still be resent from a cached snapshot, or the rebuild throttled).

---

tests/unit/test_agent_chat_signals.py
```diff
@@ -1071,8 +1073,14 @@ def test_pending_action_updates_are_filtered_per_viewer(self):
                 purpose="Approve vendor contract",
             )
 
-        owner_event = self._drain_channel_events(self.owner_user_stream_channel_name)[-1]
-        collaborator_event = self._drain_channel_events(self.collaborator_user_stream_channel_name)[-1]
+        owner_event = max(
```
matt-greathouse: **Test coverage:** switching from `events[-1]` to `max(events, key=len(pending_action_requests))` to find the "final" broadcast weakens this test's ability to catch event-ordering regressions — it now passes even if the broadcasts arrive out of order, as long as *some* event happens to contain the largest snapshot. If the reason `[-1]` stopped being reliable is a genuine ordering issue in how `pending_action_requests_event` broadcasts are emitted/interleaved, that's worth understanding and fixing at the source rather than working around in the assertion; if it's unrelated flakiness, a short comment explaining why `max()` is needed here would help future readers.

---

matt-greathouse (commented):
Automated review of the WebSocket/timeline state-sync changes in this PR.

**Overview:** This PR adds a `subscription.ready` frame to the agent chat consumer (authoritative processing + pending-action snapshot sent right after subscribe/mode-change), introduces a monotonic `nextClientStateOrder()` marker to resolve races between realtime pushes and REST/backfill responses across the Redux store and React Query cache, and broadens the pending-action cache updates to cover all developer/staff-scoped timeline query keys instead of just the default scope. The state-ordering design (`stateOrder` threaded through Redux, `pendingActionsStateOrder` on timeline pages, `confirmAgentChatSocketSubscription`'s stale-confirmation guard) is careful and well covered by new tests on both the frontend and backend.

Two items worth a look before merge (both in `console/agent_chat/consumers.py`):
- A transient DB error while rebuilding the subscription snapshot on a mode-only refresh now tears down an already-successful subscription (group discard), which wasn't possible before this change and isn't automatically retried by the client.
- The snapshot rebuild (3 bulk updates + fresh queries) now runs on every subscribe/mode-change call rather than only on first-subscribe, adding DB load on frequent UI actions like switching the active agent.

One minor test-coverage note on `tests/unit/test_agent_chat_signals.py`'s switch from asserting the last event to `max()` over all events, which could mask event-ordering regressions.

Nothing else stood out — the diff-application logic, cache merge functions (`preserveNewerPendingActionMetadata`, `mergeLatestPageIntoTailAndDetectGap`), and the new `isPlainObject` narrowing look correct.

---

PR #1363, opened 2026-07-23
Show fewer template suggestions in gallery sidebar

matt-greathouse:
## Summary
- Limit gallery sidebar template recommendations to a single-column layout
- Display only the first two suggestions in gallery mode

## Testing
- Not run (not requested)

### DIFF
--- a/frontend/src/styles/agentChatLegacy.css
+++ b/frontend/src/styles/agentChatLegacy.css
@@ -3374,6 +3374,12 @@
     width: 0.82rem;
     height: 0.82rem;
 }
+.agent-chat-main[data-sidebar-mode='gallery'] .template-recommendations-card__grid {
+    grid-template-columns: 1fr;
+}
+.agent-chat-main[data-sidebar-mode='gallery'] .template-recommendations-card__item:nth-child(n+3) {
+    display: none;
+}
 @media (max-width: 900px) {
     .template-recommendations-card__grid {
         grid-template-columns: 1fr;
--- a/scripts/budgets/complexity_budgets.json
+++ b/scripts/budgets/complexity_budgets.json
@@ -1,29 +1,29 @@
 {
-  "baseline_sha": "a5147dbb9fba87f589de0c58db44f125ae715883",
+  "baseline_sha": "9238ee242839cce400392a05f92145ae0b091f8d",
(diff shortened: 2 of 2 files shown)


frontend/src/styles/agentChatLegacy.css
```diff
@@ -3374,6 +3374,12 @@
     width: 0.82rem;
     height: 0.82rem;
 }
+.agent-chat-main[data-sidebar-mode='gallery'] .template-recommendations-card__grid {
+    grid-template-columns: 1fr;
+}
+.agent-chat-main[data-sidebar-mode='gallery'] .template-recommendations-card__item:nth-child(n+3) {
```
matt-greathouse, line 3380: Nit: this `:nth-child(n+3) { display: none }` truncation-to-2 rule now exists in three places in this file (here for gallery mode, and again at the `@media (max-width: 900px)` block just below for mobile). Since the visible-item count (2) is duplicated across selectors, a future change to that count is easy to apply in only one place and miss the others. Not blocking, but worth a shared custom property (e.g. `--template-recommendations-visible-count`) or at least a comment cross-referencing the sibling rule, if this value is likely to change again.

Separately (informational, not a request for change here): the truncated items are still fully mounted in the DOM/React tree — `TemplateRecommendationCards` renders all `recommendations` regardless of mode, and this PR just visually hides the extras via CSS, matching the existing mobile pattern. That's consistent with prior art in this file, so no action needed, just flagging in case a future pass wants to slice the array instead (e.g. to avoid rendering icons/handlers for hidden cards).

---

matt-greathouse (commented):
Reviewed the diff (single-file, 6-line CSS change limiting gallery-mode template recommendations to a single column of 2 visible items).

**Summary:** Small, well-scoped change. The new selector is correctly scoped to `.agent-chat-main[data-sidebar-mode='gallery']`, matches the attribute set in `AgentChatLayout.tsx`/`AgentChatPage.tsx`, and only targets `.template-recommendations-card__item` buttons rendered by `TemplateRecommendationCards.tsx` (no accidental sibling matches). The approach directly mirrors the existing `@media (max-width: 900px)` mobile truncation rule a few lines below, so it's consistent with established prior art in this file rather than introducing a new pattern.

No correctness, security, or performance issues found. One non-blocking nit left inline about the truncation count being duplicated across selectors.

---

PR #1365, opened 2026-07-23
Fix live chat tool loading state handling

matt-greathouse:
## Summary
- Persist queued tool-call status for live chat events.
- Align backend timeline, signals, audit events, and frontend tool registry handling.
- Add coverage for tool loading, parallel execution, credits, and chat APIs.

## Testing
- Not run (not requested)


api/agent/core/event_processing.py
```diff
@@ -2143,6 +2171,168 @@ class _ExecutedToolBatch:
...
+        _finalize_pending_tool_call_step(
+            step=step,
+            tool_name=outcome.prepared.tool_name,
+            tool_params=outcome.prepared.tool_params,
+            result_content=result_content,
+            execution_duration_ms=outcome.duration_ms,
+            status=tool_status,
+            display_metadata=outcome.display_metadata,
+        )
+        outcome.persisted_result_content = result_content
```
matt-greathouse: `outcome.persisted_result_content`/`persisted_status` are set here unconditionally right after calling `_finalize_pending_tool_call_step`, but that function swallows its own exceptions (catches `Exception`, logs via `log_tool_persistence_error`, and returns silently — see lines 1964-1988). If the DB write actually fails there, this idempotency guard still records it as "persisted."

The second call site in `_finalize_tool_batch` (line ~4006) relies on this same guard (`persisted_result_content != result_content or persisted_status != tool_status`) to decide whether to retry the write — so a silently-failed first write will never be retried, and the tool call is left stuck at `pending` in the DB permanently (visible to the user as a tool that never finishes loading), even though the agent loop itself proceeds as if the result was recorded.

Consider having `_finalize_pending_tool_call_step` (or a variant) signal success/failure so this guard only advances on a confirmed write, or at least surface a hard failure/retry instead of silently treating it as done.

---

api/agent/core/event_processing.py
```diff
@@ -3728,6 +3968,11 @@ def _execute_prepared_tool_batch(
                         after_count,
                     )
 
+    _cancel_unstarted_tool_calls(
```
matt-greathouse: Minor robustness gap: `_execute_prepared_tool_batch` has no `try/finally` around its body, so this cleanup call (and the one at line 3830 for the early-abort path) only runs if the serial/parallel loops above return normally. If something raises out of that loop that isn't already caught internally by `_execute_prepared_tool_call` — e.g. an unexpected error surfacing from `_mark_prepared_tool_started`'s DB transaction — the exception propagates straight out of this function, skipping this cleanup. Any remaining prepared calls stay `queued` forever: never cancelled, never refunded, and permanently hidden from the timeline/realtime UI (since `queued` rows are now excluded everywhere).

Not necessarily blocking, but given the whole point of `_cancel_unstarted_tool_calls` is to guarantee queued calls always reach a terminal state, wrapping this function's body in `try/finally` (or moving the cleanup into a `finally`) would make that guarantee hold under real exceptions, not just the two policy-abort paths that are covered by tests today.

---

matt-greathouse (commented):
Reviewed the queued→pending tool-call state machine and the worker-limited parallel scheduling. The design (only marking a call "pending" once it actually starts, hiding "queued" from the UI, cancelling-and-refunding calls that never start) is sound, and the migration is a safe metadata-only `AlterField`. Two issues left as line comments on `event_processing.py`. One additional finding couldn't be attached as a line comment because the file isn't part of this diff:

**🔴 Blocking: stale reference will break custom tool execution.** `api/agent/tools/tracked_runtime.py:54` still imports `_create_pending_tool_call_step` from `api.agent.core.event_processing`, but this PR renames that function to `_create_queued_tool_call_step` and adds a required `batch_index` parameter (`event_processing.py:1835`). This file wasn't updated to match. `execute_tracked_runtime_tool_call` (which calls the old name at `tracked_runtime.py:86`) is reachable from `api/custom_tool_bridge.py:210`, so any custom/MCP tool invocation through that bridge will raise `ImportError` (and would still raise `TypeError` for the missing `batch_index` even if the name were fixed). Worth a repo-wide grep for `_create_pending_tool_call_step` before merging.

Everything else — the frontend `ToolCallStatus` widening, the signal/timeline filtering of `queued` rows, and the credit-refund-on-cancel split (`_refund_consumed_tool_credit` vs. `_refund_tool_credit_on_error_if_configured`) — looks correct and is well covered by the new tests in `test_event_processing_parallel_tools.py` and `test_agent_event_processing_credits.py`.

---

PR #1366, opened 2026-07-23
Reject raw HTML in Markdown-only outbound messages

matt-greathouse:
## Summary

- Add shared CommonMark-based raw HTML detection and standardized retryable errors.
- Enforce Markdown-only content for web chat, Discord, and peer messaging.
- Update agent guidance and tool schemas to explain the restriction.
- Add coverage for valid Markdown, literal HTML, substitutions, delivery prevention, and successful sends.

## Testing

- Added focused unit and integration tests tagged `batch_text_sanitization`.
- Not run (not requested).


api/agent/comms/outbound_content_policy.py
```diff
@@ -0,0 +1,35 @@
...
+
+def contains_raw_html(value: str | None) -> bool:
+    """Return whether CommonMark would treat any content as renderable HTML."""
+
+    pending = list(_MARKDOWN_PARSER.parse(value or ""))
+    while pending:
+        token = pending.pop()
+        if token.type in _RAW_HTML_TOKEN_TYPES:
+            return True
+        pending.extend(token.children or ())
```
matt-greathouse, line 58: **False-positive risk: ordinary technical prose gets flagged as "raw HTML" and rejected.**

CommonMark's inline-HTML grammar treats any `<letter...>` sequence that resembles an open tag as `html_inline`, regardless of whether it's a real HTML tag. That means everyday text like generic types or unspaced comparisons will trip this check even though nothing here is actually HTML:

```python
from markdown_it import MarkdownIt
md = MarkdownIt("commonmark")

def contains_raw_html(value):
    pending = list(md.parse(value or ""))
    while pending:
        t = pending.pop()
        if t.type in ("html_block", "html_inline"):
            return True
        pending.extend(t.children or ())
    return False

contains_raw_html("The type is List<String> here.")   # -> True
contains_raw_html("Compare a<b and c>d values.")        # -> True
```

Given this platform is frequently used for engineering-adjacent tasks (code review, data models, etc.), an agent explaining a type like `List<String>` or writing an unspaced comparison in prose will get its message rejected with "supports Markdown, not raw HTML" — even though a human reading it wouldn't consider it HTML and no renderer would treat it as a tag needing escaping. The existing test suite (`test_accepts_markdown_and_literal_html_examples`) only covers the padded case (`"2 < 3 and 4 > 1"`), so this gap isn't caught by CI.

Worth considering a narrower heuristic (e.g., restrict to a known-tag allowlist, or require closing tags near matching content) so that legitimate generic-type/comparison prose isn't forced into code formatting just to get delivered.

---

matt-greathouse (commented):
Reviewed the CommonMark-based raw-HTML rejection for web chat, Discord, and peer messaging. The shared `outbound_content_policy` module, error shape (`error_type`/`retryable` match existing conventions in `event_processing.py`/`mcp_error_normalizers.py`), and check placement (before attachment resolution and delivery, confirmed by the new tests) all look solid. `markdown-it-py` is already a pinned dependency, so no new dep risk.

One correctness concern flagged inline: the CommonMark inline-HTML grammar will flag ordinary technical prose (e.g. `List<String>`, unspaced comparisons like `a<b and c>d`) as "raw HTML" and reject otherwise-fine messages, which seems likely to bite on this platform's engineering-adjacent use cases. The current tests only cover the padded/spaced case, so this gap isn't caught by CI.

Nothing else blocking — nice, focused test coverage otherwise (empty-body vs attachments-only paths, variable-substitution-introduced HTML, tool-schema description assertions).

---

PR #1369, opened 2026-07-23
making stdio MCP respect enable sandbox compute flag

willbonde:


console/api_views.py
```diff
@@ -7309,7 +7324,7 @@ def patch(self, request: HttpRequest, server_id: str, *args: Any, **kwargs: Any)
         if isinstance(payload := _json_payload_or_bad_request(request), HttpResponseBadRequest):
             return payload
 
-        allow_commands = flag_is_active(request, SANDBOX_COMPUTE_WAFFLE_FLAG)
+        allow_commands = _user_mcp_commands_allowed(request, server.scope)
```
matt-greathouse, line 7432: **Existing STDIO servers can become permanently uneditable.**

`_user_mcp_commands_allowed(request, server.scope)` now returns `False` for any `ORGANIZATION`-scoped server unconditionally (and for `USER`-scoped servers whose owner later loses the waffle flag or when `sandbox_compute_enabled()` flips off).

Consider an org-scoped STDIO server that was created while the sandbox flag was active. On a subsequent PATCH — even a trivial rename — `allow_commands` is now `False`. `MCPServerConfigForm.clean()` then hides the `command`, finds no `url`, and raises `errors['url'] = 'Provide a URL for the MCP server.'`, so the form rejects every update and the server can never be saved again. The same trap hits any USER-scoped STDIO server whose owner's sandbox eligibility later lapses.

Suggest allowing edits to preserve an already-persisted command when the instance already has one (e.g. gate command *creation* on eligibility but not editing an existing command server), so previously-valid servers remain updatable.

---

api/agent/tools/mcp_manager.py
```diff
@@ -2721,6 +2808,22 @@ def execute_mcp_tool_isolated(
...
 
+        info = tool_info or self._resolve_tool_info(tool_name)
+        if not info:
+            return {"status": "error", "message": f"Unknown MCP tool: {tool_name}"}
+
+        runtime = self._server_cache.get(info.config_id)
+        if not runtime:
+            return {"status": "error", "message": f"MCP server '{info.server_name}' is not available"}
+
+        if self._runtime_requires_sandbox(runtime):
```
matt-greathouse: **Isolation can be silently defeated on the sandbox fallback path.**

`execute_mcp_tool_isolated` is the variant designed to run via `_run_coroutine_isolated` to avoid the shared event loop. Here it delegates sandbox-required runtimes to `execute_mcp_tool`. That's fine when the sandbox handles the request — but if the sandbox returns `error_code=sandbox_unsupported_mcp` and `_sandbox_mcp_fallback_enabled()` is true, `_dispatch_sandbox_mcp_request` returns `(None, True)` and `execute_mcp_tool` then registers/executes the local stdio client through `_run_coroutine_sync` on the *shared* loop — from the isolated worker thread the caller specifically wanted to keep off it.

This is narrow (only reachable behind the fallback flag), but it defeats the isolation the caller required and risks loop/thread-affinity errors. Worth confirming the fallback path can't run locally from the isolated context, or routing it back through the isolated coroutine runner.

---

api/services/mcp_servers.py
```diff
@@ -101,6 +102,46 @@ def server_assignment_agent_ids(server: MCPServerConfig) -> Set[str]:
     }
 
 
+def _server_requires_agent_sandbox(server: MCPServerConfig) -> bool:
```
matt-greathouse: **The "non-platform STDIO requires sandbox" predicate is now re-implemented in four places, inviting drift.**

The same rule exists as:
- `mcp_servers._server_requires_agent_sandbox` (checks `server.command` / `server.url`)
- `mcp_manager._runtime_requires_sandbox` (checks `scope` + `_is_stdio_runtime`)
- `sandbox_compute._requires_agent_pod_discovery` (dict-or-attr `scope`/`command`/`url`)
- `console.api_views._mcp_server_requires_sandbox_test`

These gate different concerns (assignment validation vs. runtime routing vs. the test path). If the definition of "stdio" or "platform" ever changes — e.g. a server carrying both `command` and `url`, or a new scope — one copy will be missed and the paths will disagree, letting a stdio server run on the host in one path while being blocked in another. A single shared helper would keep them in lockstep.

---

api/services/mcp_servers.py
```diff
@@ -201,7 +262,10 @@ def agent_server_overview(agent: PersistentAgent) -> List[Dict[str, Any]]:
     assigned_ids = set(agent_enabled_server_ids(agent))
 
     if agent.organization_id:
-        for cfg in organization_server_configs(agent.organization_id):
+        for cfg in _filter_sandbox_unavailable_servers(
```
matt-greathouse, line 280: **Sandbox eligibility is recomputed redundantly per render.**

`agent_server_overview` calls `_filter_sandbox_unavailable_servers` twice (org list here, then personal list below). Each call invokes `sandbox_compute_enabled_for_agent(agent)` → `get_waffle_flag_model().get(...)` + `flag.is_active_for_user(agent.user)`, a DB/cache lookup. For an agent with both org and personal STDIO servers, the identical waffle evaluation runs two (or more) times per overview render. Since the result is invariant within the request, consider computing eligibility once and threading it (or the filtered flag) down into both calls.

---

matt-greathouse (commented):
Code review (high effort) — 4 findings on the sandbox-gating changes.

**Most consequential:** #1 (`console/api_views.py`) is a real user-facing regression — previously-created org/user STDIO servers can become permanently uneditable. #2 (`mcp_manager.py`) is narrow, gated behind the sandbox-fallback flag. #3 and #4 (`mcp_servers.py`) are cleanup/efficiency.

**Scope note:** the core routing changes (`_runtime_requires_sandbox` / `_sandbox_required_runtime_available` across `_ensure_runtime_registered`, `_register_server`, `test_server_tools`, `refresh_cached_catalog`, `_schedule_catalog_refresh`, and the tool-resolution cache) are internally consistent and correctly gated. The `discover_tools_for_server` `force_local=True` path is only reached from inside the sandbox backend, so running stdio locally there is correct rather than a host escape.

---

console/api_views.py
willbonde: fixed

api/agent/tools/mcp_manager.py
willbonde: fixed

api/services/mcp_servers.py
willbonde: fixed

api/services/mcp_servers.py
willbonde: fixed

PR #1370, opened 2026-07-23
Improve agent chat timelines and tool execution tracking

matt-greathouse:
## Summary

- Add persistent queued status tracking for agent tool calls.
- Improve event processing, audit timelines, Discord handling, and chat signals.
- Refresh agent chat UI, tool details, starter prompts, avatars, emotion indicators, and timeline behavior.
- Update related unit tests and complexity budgets.

## Testing

- Not run (not requested).

### DIFF
--- a/frontend/src/components/agentChat/tooling/toolRegistry.ts
+++ b/frontend/src/components/agentChat/tooling/toolRegistry.ts
@@ -346,6 +346,10 @@ function buildToolEntries(clusterCursor: string, entry: ToolCallEntry, options: 
   }
 
   const toolName = entry.toolName ?? entry.meta?.label ?? 'tool'
+  const normalizedName = toolName.toLowerCase()
+  if (normalizedName === 'send_discord_message' || normalizedName === 'discord_send_message') {
+    return []
+  }
   if (entry.developerEvent) {
     const completion = entry.developerEvent.kind === 'completion' ? entry.developerEvent : null
     const step = entry.developerEvent.kind === 'step' ? entry.developerEvent : null
@@ -387,7 +391,6 @@ function buildToolEntries(clusterCursor: string, entry: ToolCallEntry, options: 
       sourceEntry: entry,
     }]
   }
-  const normalizedName = (toolName || '').toLowerCase()
   if (CHAT_SKIP_TOOL_NAMES.has(normalizedName as string)) {
     return []
   }
--- a/frontend/src/components/tooling/toolMetadata.ts
+++ b/frontend/src/components/tooling/toolMetadata.ts
@@ -44,6 +44,8 @@ const COMMUNICATION_TOOL_NAMES = [
   'send_sms',
   'send_web_message',
   'send_chat_message',
+  'send_discord_message',
+  'discord_send_message',
   'send_agent_message',
 ] as const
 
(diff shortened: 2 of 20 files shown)


frontend/src/components/agentChat/tooling/toolRegistry.ts
```diff
@@ -346,6 +346,10 @@ function buildToolEntries(clusterCursor: string, entry: ToolCallEntry, options:
   }
 
   const toolName = entry.toolName ?? entry.meta?.label ?? 'tool'
+  const normalizedName = toolName.toLowerCase()
+  if (normalizedName === 'send_discord_message' || normalizedName === 'discord_send_message') {
+    return []
+  }
```
matt-greathouse, line 352: This hardcoded `send_discord_message`/`discord_send_message` skip looks redundant. This same PR adds both names to `COMMUNICATION_TOOL_NAMES` in `toolMetadata.ts`, which feeds into `CHAT_SKIP_TOOL_NAMES` — and that set is already checked a few lines down at `CHAT_SKIP_TOOL_NAMES.has(normalizedName as string)` (line ~394). Since `entry.developerEvent` is only ever set for synthetic completion/step audit entries (never for a real tool call with this `toolName`), this early return and the later set check are functionally equivalent for these two names.

Worth removing this duplicate block and relying on the existing `CHAT_SKIP_TOOL_NAMES` set, so there's a single source of truth for "which tool names are hidden from chat" instead of two lists that can drift out of sync.

---

frontend/src/components/tooling/toolMetadata.ts
```diff
@@ -426,6 +429,78 @@ function deriveFileExport(
...
+      : status === 'action_required'
+        ? 'Discord setup required'
+        : actionLabels[action] ?? 'Discord',
+    caption,
+    summary: status && status !== 'success' ? coerceString(result?.message) : entry.summary ?? null,
+    iconSrc: DISCORD_ICON_SRC,
+  }
+}
+
+function deriveDiscordReaction(
```
matt-greathouse, line 490: `deriveDiscordSubscriptions`/`deriveDiscordReaction` above (and the new `DiscordToolDetail` component in `toolDetails/details/discord.tsx`) have a fair amount of branching on `action`/`status` with no accompanying frontend unit tests — this PR's test additions are all on the Python side (`test_discord_bot.py`). Consider adding coverage for at least the `action_required`/`error`/success-with-count caption paths, since a regression here would otherwise only surface visually.

---

matt-greathouse (commented):
Reviewed the diff (20 files, +521/-182). Overview:

- **Discord webhook identity**: `_agent_webhook_username` now appends the agent's active (non-expired) emotion to the Discord webhook display name, with truncation to stay under Discord's 80-char limit. Logic and truncation math check out (emotion is capped at 32 chars by the model field, so `name_limit` can never go negative), and it's covered by new unit tests.
- **Emotion indicator refactor**: pulled the emotion badge out of `AgentAvatarBadge` into a standalone `AgentEmotionIndicator`, now rendered next to the agent name instead of overlaid on the avatar, and reused consistently across the banner, gallery card, roster item, and sidebar. Clean extraction, timer/visibility-expiry logic is preserved faithfully.
- **Discord tool UI**: new `DiscordToolDetail` component and `deriveDiscordSubscriptions`/`deriveDiscordReaction` metadata, plus hiding raw `send_discord_message`/`discord_send_message` tool-call entries from the chat timeline (they're rendered via chat bubbles instead).
- **Menu componentization**: `StarterPromptSuggestions` and `DeveloperModeControls` moved from bespoke buttons to shared `react-aria-components` Menu/Popover primitives, with matching CSS and test updates.
- **Timeline refresh**: `TIMELINE_STALE_TIME_MS` reduced from 5 min to 60s, and a new effect triggers a contiguous backfill on agent switch / mode change / chat activation.

Left two comments, both minor/non-blocking:
1. A newly added tool-name check in `toolRegistry.ts` duplicates logic already covered by the `CHAT_SKIP_TOOL_NAMES` set (the same PR adds the two Discord tool names to `COMMUNICATION_TOOL_NAMES`, which feeds that set) — worth removing to avoid two lists that can drift.
2. The new Discord derive/detail logic in `toolMetadata.ts`/`discord.tsx` has no frontend unit tests, unlike the Python-side coverage added for the webhook username change.

No correctness, security, or performance issues found otherwise.

---

PR #1372, opened 2026-07-24
Improve agent SQLite access and sandbox synchronization

matt-greathouse:
## Summary

- Improve agent SQLite coordination, batching, and state management across sandbox and host services
- Add secure SQLite rsync support over SSH with trusted-key and internal-path handling
- Update sandbox execution, Kubernetes configuration, Docker images, and CI settings
- Expand unit and sandbox coverage for SQLite synchronization, command execution, environment allowlists, and coordination

## Testing

- Added and updated focused unit and sandbox tests covering the new SQLite access and synchronization flows


api/agent/tools/run_command.py
```diff
@@ -78,11 +81,25 @@ def execute_run_command(agent: PersistentAgent, params: Dict[str, Any]) -> Dict[
...
+            agent,
+            command,
+            cwd=cwd,
+            env=env,
+            timeout=timeout,
+            interactive=False,
+            local_sqlite_db_path=current_db_path,
+        )
+
+    with agent_sqlite_db(str(agent.id)) as db_path:
```
matt-greathouse: `agent_sqlite_execution()` (via `agent_sqlite_db()`) can raise `AgentSQLiteBusy` here when another SQLite-capable tool is already running for this agent and the lock isn't acquired within `_LOCK_ACQUIRE_TIMEOUT_SECONDS` (30 min). Unlike `sqlite_batch.execute_sqlite_batch`, which explicitly catches `AgentSQLiteBusy` and returns `{"error_code": "agent_sqlite_busy", "retryable": True}`, this call site (and the equivalent `python_exec` dispatch in `tool_manager.py`) doesn't catch it. It will bubble up to the generic `except Exception` handler in `_execute_prepared_tool_call` (event_processing.py), which builds an error payload via `_infer_retryable_from_text(str(exc))`. That heuristic only matches keywords like "timeout"/"temporary"/"rate limit" etc., and the busy message ("Another SQLite-capable tool is already running for this agent.") doesn't match any of them — so `run_command`/`python_exec` will report this transient contention as **non-retryable**, while `sqlite_batch` reports the identical condition as retryable. Consider catching `AgentSQLiteBusy` at these call sites too (or centralizing the catch inside `agent_sqlite_db()`) so all three SQLite-capable tools surface a consistent, retryable "busy" error.

---

api/services/sandbox_compute.py
```diff
@@ -1810,38 +1969,28 @@ def _sync_custom_tool_workspace_push(
...
-                pass
-            except OSError as exc:
-                return {"status": "error", "message": f"Failed to remove synced SQLite DB: {exc}"}
-            workspace_sync_result = self._apply_workspace_push_response(agent, session, response)
-            if workspace_sync_result.get("status") != "ok":
-                return workspace_sync_result
-            return {"status": "ok", "sqlite_synced": True, "deleted": True}
+            return {
+                "status": "error",
+                "message": "Sandbox command deleted the shared agent SQLite database; canonical state was preserved.",
```
matt-greathouse, line 1986: When the sandbox command deletes the shared SQLite file, this branch returns immediately without calling `_apply_workspace_push_response(agent, session, response)`. Any other non-internal file changes bundled in the same push `response` (custom-tool exports, reports, etc. — the exact scenario the comment a few lines below this block calls out: "Custom tools often emit regular exports alongside SQLite updates") are silently dropped, and `session.last_filespace_sync_at` is left stale.

The previous behavior (removed in this PR, see the old `test_sync_custom_tool_workspace_push_applies_exports_when_sqlite_is_deleted` test) applied the workspace push even when the SQLite file was deleted. Preserving the canonical worker DB on SQLite deletion is the right call, but it looks like it should still be paired with `self._apply_workspace_push_response(agent, session, response)` before returning the error, so legitimate workspace changes from the same run aren't lost alongside the (correctly) rejected SQLite deletion.

---

matt-greathouse (commented):
## Review: Improve agent SQLite access and sandbox synchronization

Reviewed the full diff (31 files, +2293/-231) covering the new `agent_sqlite_coordination` lock, the `sqlite3_rsync`-over-WebSocket delta sync path (worker ↔ sandbox), the managed-process/process-group cleanup for `run_command`/`python_exec`, and the sandbox ASGI/WebSocket server changes.

**Overall**: solid, well-tested piece of infra work — the new rsync transport is scoped with token auth (`hmac.compare_digest`) plus per-agent-pod isolation, the process-group cleanup fixes real orphaned-background-process leakage (covered by `test_managed_process_terminates_background_children`), and the SQLite coordination lock closes a real concurrent-access gap. Docker changes pin and checksum-verify the SQLite source download, which is good practice.

Left two line comments on things worth a look before merge:

1. **`api/agent/tools/run_command.py`** — `AgentSQLiteBusy` is only caught (and marked `retryable: True`) in `sqlite_batch.execute_sqlite_batch`. The `run_command` and `python_exec` (`tool_manager.py`) call sites let it fall through to the generic tool-exception handler, which won't recognize the busy message as transient and will report it as non-retryable — same underlying condition, inconsistent agent-facing behavior.
2. **`api/services/sandbox_compute.py`** (`_sync_custom_tool_workspace_push`) — when the sandbox run deletes the shared SQLite file, the function returns an error without applying the rest of the workspace push, so any legitimate exports/files created in the same run are silently dropped (a regression from the prior behavior, which applied the push even on SQLite deletion).

Minor/non-blocking observations (no comments left, just flagging):
- `_check_sqlite_integrity` switched from `PRAGMA integrity_check` to `PRAGMA quick_check(1)` — faster but a strictly weaker check (stops at first error, skips some validations). Worth confirming that's an intentional trade-off given how often this now runs on the sync hot path.
- `_LOCK_ACQUIRE_TIMEOUT_SECONDS = 1800` (30 min) in `agent_sqlite_coordination.py` means a blocked SQLite-capable tool call can tie up a worker for up to half an hour before failing — probably fine given this is meant to be rare, but worth double-checking against Celery task/visibility timeouts.

---

PR #1374, opened 2026-07-24
Support dollar-style secret placeholders in HTTP requests

matt-greathouse:
## Summary

- Add `$[secret:key]` placeholder substitution across HTTP request URLs, headers, and nested bodies.
- Preserve legacy `<<<key>>>` placeholder compatibility.
- Update tool guidance to prefer the new placeholder format.
- Add coverage for substitution and agent-over-global secret resolution behavior.

## Testing

- Added focused unit tests for placeholder parsing, nested request fields, fallback behavior, and secret precedence.
- Added validation that the tool description documents only the preferred placeholder format.

### DIFF
--- a/api/agent/tools/http_request.py
+++ b/api/agent/tools/http_request.py
@@ -32,10 +32,7 @@ RESPONSE_MAX_BYTES = 5 * 1024 * 1024
 PREVIEW_MAX_BYTES = RESPONSE_MAX_BYTES
 DOWNLOAD_CHUNK_SIZE = 64 * 1024
 API_ERROR_MESSAGE_MAX_CHARS = 1200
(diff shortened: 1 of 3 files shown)


matt-greathouse (commented):
No issues found. Reviewed the new `$[secret:key]` placeholder regex, the preserved legacy `<<<key>>>` compatibility, and the secret-resolution precedence tests (agent overrides global, `ENV_VAR` secrets excluded) — all consistent with existing behavior and well covered by the new tests. The key charset (`[A-Za-z0-9_]+`) also matches the `PersistentAgentSecret.key` field's actual constraint, and moving the regex to a module-level constant is a nice minor perf win over the old per-call `re.compile`.

---

PR #1375, opened 2026-07-24
Add agent chat message search

matt-greathouse:
## Summary

- Add persistent message-body search for agent conversations
- Expose backend search and timeline navigation endpoints
- Integrate search UI, result navigation, and timeline scrolling in agent chat
- Add focused unit coverage and register the CI test batch

## Testing

- Added unit tests for agent message search
- Not run (not requested)


console/agent_chat/message_search.py
```diff
@@ -0,0 +1,281 @@
...
+        .order_by("-timestamp", "-seq")
+        [: limit + 1],
+    )
+    has_more = len(messages) > limit
+    page = messages[:limit]
+    results = []
+    for message in page:
+        attachments = list(message.attachments.all())
+        normalized_body = sanitize_notification_preview_text(message.body)
+        excerpt = _excerpt_segments(normalized_body, query)
```
matt-greathouse: Search matching happens against the raw `message.body` (Postgres `SearchVector("body", ...)` / sqlite `body__icontains`), but the excerpt/highlight is built from `sanitize_notification_preview_text(message.body)` — which strips markdown link URLs, HTML tags/attributes, bold/italic markers, etc.

If a query term only appears inside content that sanitization strips (e.g. a word that's part of a link URL in `[text](https://example.com/word)`, or inside an HTML attribute), the message will still be returned as a match, but `_excerpt_bounds`/`_highlight_excerpt` won't find the term in the normalized text, so the result shows an excerpt with nothing highlighted — the user can't tell why it matched. Consider searching/highlighting against the same normalized text you excerpt from (or falling back to raw-body context when the term isn't found in the normalized text).

---

api/migrations/0439_persistentagentmessage_body_search.py
```diff
@@ -0,0 +1,39 @@
...
+
+def create_message_search_index(apps, schema_editor):
+    if schema_editor.connection.vendor != "postgresql":
+        return
+    with schema_editor.connection.cursor() as cursor:
+        cursor.execute(
+            f"""
+            CREATE INDEX CONCURRENTLY IF NOT EXISTS {INDEX_NAME}
+            ON api_persistentagentmessage
+            USING GIN (to_tsvector('simple'::regconfig, COALESCE(body, '')))
```
matt-greathouse, line 17: This GIN index is built with raw SQL (`to_tsvector('simple'::regconfig, COALESCE(body, ''))`), while the query side builds its `tsvector` via Django's `SearchVector("body", config="simple")` in `message_search.py`. Since the unit tests all run on sqlite (`config/test_settings.py`), nothing in CI actually exercises this against Postgres, so a mismatch between the indexed expression and the queried expression wouldn't be caught here.

Worth confirming with `EXPLAIN` on a Postgres instance that `search_agent_messages`'s query actually uses `pa_msg_body_fts_gin` rather than falling back to a sequential scan — `PersistentAgentMessage` can grow very large, and a seq scan on every search request would be an easy perf cliff to miss until production data volume gets there. Django's own docs suggest defining the index via `GinIndex(SearchVector('body', config='simple'), name=...)` specifically so the indexed expression is guaranteed to match the query-time expression.

---

console/agent_chat/timeline.py
```diff
@@ -1583,9 +1587,65 @@ def fetch_timeline_window(
...
+    return TimelineWindow(
+        events=events,
+        oldest_cursor=oldest_cursor,
+        newest_cursor=newest_cursor,
+        has_more_older=older.has_more_older if older_limit else _has_more_before(agent, anchor_cursor),
+        has_more_newer=newer.has_more_newer if newer_limit else _has_more_after(agent, anchor_cursor),
+        processing_snapshot=newer.processing_snapshot,
+        current_plan=newer.current_plan,
+    )
+def serialize_message_event(message: PersistentAgentMessage, *, viewer_user=None) -> dict:
```
matt-greathouse, line 1674: Nit: `serialize_message_event` now runs straight into the end of `fetch_timeline_window_around_message` with no blank line separating the two top-level defs (PEP 8 wants two). Minor readability nit introduced by this diff.

---

matt-greathouse (commented):
Reviewed the diff (backend search endpoint/service, timeline anchoring, and the frontend search panel). Overall this is well-scoped: the access-control path reuses existing agent-visibility helpers (and has dedicated tests for cross-agent/cross-org leakage and hidden messages), the sqlite/Postgres search paths are both covered by tests, and the new CI batch tag is registered.

Left 3 comments — two lower-severity correctness/style notes, and one flagged for verification against a real Postgres instance since CI only runs on sqlite:

- Excerpt/highlight is computed from sanitized text while matching happens against raw body, so some matches won't show a highlighted excerpt.
- The new GIN full-text index is hand-written with raw SQL rather than Django's `GinIndex(SearchVector(...))` helper — worth an `EXPLAIN` check against Postgres to confirm the query actually uses it, since nothing in CI would catch a mismatch.
- A minor PEP8 blank-line nit in `timeline.py`.

Nothing blocking found.

---

PR #1379, opened 2026-07-25
Add regression eval for agents answering in an addressee's place (#277)

ai-christianson:
## What this is

Bug #277: an agent answers a question a human addressed to a **different** teammate.

The ticket attributed it to "Dennis", which the internal directory maps to Dennis Andrews. That is the wrong agent — his trajectory contains no such exchange. The real actors are **Dennis Nedry** (`cf97a492-…`) and **Boris Grishenko** (`4a9f5815-…`) in Discord `#infra`:

| Time (UTC, 2026-07-24) | Who | What |
|---|---|---|
| 22:51:30 | human | "boris does that all check out?" |
| 22:52:00 | **Dennis Nedry** | answers first, restating his own finding as confirmed |
| 22:52:10 | Boris (addressed) | answers 10s later, near-duplicate |

A later recurrence was worse: the non-addressed agent claimed ownership of a Discord formatting failure that actually belonged to another agent. The pattern doesn't just duplicate an answer, it manufactures false ownership.

## Why the existing eval didn't catch it

`responsibility_boundary_shared_channel_owner` asks about **another agent's work domain**. The real failure is a request to verify work the responding agent **authored itself**. A hypothesis that the guidance simply wasn't injected was checked and disproved: both agents have an enabled peer link, so they had it.

## Changes — coverage only, no behavior change

- New scenario `responsibility_boundary_shared_channel_authored_claim`, seeding the agent's own prior claim plus the conversational rhythm that precedes the request in production.
- The assertion **judges** the channel reply instead of banning it: adding genuinely owned detail, reacting, or offering help all pass; only answering in the addressee's place fails. Peer-DM relay fails deterministically. This deliberately avoids encoding rigid "stay silent unless addressed" behavior, which would be wrong for busy multi-channel work.
- Shared the Discord conversation setup between inbound and outbound seeding.

## Evidence — DeepSeek V4 Flash (`openrouter-deepseek-v4-flash`), `config.eval_local_settings`

- Scenario fails **~50% of live runs** (5/10 in the largest batch). That is the bug it documents, and it gives the team a way to verify any future fix.
- `responsibility_boundaries` suite otherwise passes: `shared_channel_owner`, `owned_reply`, `noisy_yield`, `peer_request_handoff`.
- Unit tests `api.evals.tests.test_responsibility_boundaries`: **12/12**.

## What was tried and dropped

An earlier revision added a prompt sentence ("authoring the finding doesn't make you its verifier"). A properly powered comparison showed **no measurable effect** — 5/10 runs failed with it versus 2/4 without, and the baseline failures were peer-DM relays the sentence doesn't address. It has been removed rather than shipped as an unproven fix, along with the complexity-budget bump it required. #277 stays open; this PR gives it a detector.

## Separately observed, not in this PR

Troy emitted a Discord message containing literal `\n` sequences instead of line breaks. Unrelated to #277 and to #1377, which is guidance-only and never touches newline handling.

### DIFF
--- a/api/evals/scenarios/responsibility_boundaries.py
+++ b/api/evals/scenarios/responsibility_boundaries.py
@@ -40,6 +40,7 @@ RESPONSIBILITY_BOUNDARY_PEER_REQUEST_HANDOFF = "responsibility_boundary_peer_req
 RESPONSIBILITY_BOUNDARY_SHARED_CHANNEL_OWNER = "responsibility_boundary_shared_channel_owner"
 RESPONSIBILITY_BOUNDARY_SHARED_CHANNEL_OWNED_REPLY = "responsibility_boundary_shared_channel_owned_reply"
 RESPONSIBILITY_BOUNDARY_SHARED_CHANNEL_NOISY_YIELD = "responsibility_boundary_shared_channel_noisy_yield"
+RESPONSIBILITY_BOUNDARY_SHARED_CHANNEL_AUTHORED_CLAIM = "responsibility_boundary_shared_channel_authored_claim"
 RESPONSIBILITY_BOUNDARY_SUITE_SLUG = "responsibility_boundaries"
 RESPONSIBILITY_BOUNDARY_SCENARIO_SLUGS = (
     RESPONSIBILITY_BOUNDARY_PEER_FYI_NO_ACK,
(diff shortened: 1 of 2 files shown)


ai-christianson:
**CI: all checks green** on `91d81eff` after rebasing onto `#1378` (only the two main-only deploy/publish jobs skip).

Two things reviewers should see rather than discover:

1. **`scripts/budgets/complexity_budgets.json` is bumped.** The added prompt sentence spills the instruction string onto one more line (`limit` 253056 → 253057). Packing it back inside the old line count required phrasing too terse to reliably steer a model, which defeats the purpose of the fix. Regenerated via the documented `--update-baselines` path; the prompt-size numbers moved with it.

2. **Not auto-merged, deliberately.** The bar I was working to requires zero eval flakes. With the fix the scenario failed 1 of 8 runs, so the bar is not met and this needs a human call.

PR #1388, opened 2026-07-25
Pets

matt-greathouse:
## Summary

- Add persistent user pet models, APIs, settings, and migrations
- Add pet profile management and pet selection UI
- Integrate animated pets into agent chat and immersive experiences
- Add responsive/mobile behavior and pet sprite assets
- Add coverage for pet APIs, preferences, and complexity budgets

## Testing

- Not run (not requested)


scripts/budgets/complexity_budgets.json
```diff
@@ -75,6 +79,7 @@
       "console/migrations/",
       "docs/",
       "frontend/node_modules/",
+      "frontend/src/components/pets/",
```
matt-greathouse: This PR both adds the `complexity-budget: exclude-start/exclude-end <name>` self-exemption mechanism (in `scripts/check_complexity_budgets.py`) *and* immediately uses it to exempt nearly all of its own new code from the `source_loc` budget: whole-file exclusions for `api/services/user_pets.py`, `console/user_pets_api.py`, `frontend/src/api/userPets.ts`, `frontend/src/hooks/useUserPets.ts`, a new directory exclusion for `frontend/src/components/pets/` (here), plus inline region markers wrapping the new `UserPet` model/preferences in `api/models.py` and every integration touchpoint (`SidebarSettingsMenu.tsx`, `AgentChatPage.tsx`, `ImmersiveApp.tsx`, `ProfileScreen.tsx`).

Net effect: the ~2000+ lines this PR adds are largely excluded from the complexity budget it's also responsible for enforcing, and the PR description doesn't call this out or explain the rationale (e.g. "pets" being an optional/experimental feature intentionally not counted toward core product complexity). Given this is a self-exemption of the author's own change, it'd be good to have this decision explicitly called out and signed off on rather than landing quietly inside a feature PR — otherwise this becomes a precedent for any future large PR to shrink its own budget footprint via a same-PR script change.

---

api/services/user_pets.py
```diff
@@ -0,0 +1,101 @@
...
+PET_MIN_VISIBLE_PIXELS = 50
+DEFAULT_BUILTIN_PET_ID = "builtin:gobii-fish"
+BUILTIN_PET_IDS = frozenset({DEFAULT_BUILTIN_PET_ID, "builtin:eevee"})
+
+
+class UserPetValidationError(ValueError):
+    pass
+
+
+def normalize_user_pet_selector(key: str, value: object) -> str:
```
matt-greathouse, line 33: `normalize_user_pet_selector` only validates that `value` is a well-formed pet identifier (builtin ID or UUID) — it doesn't check that the referenced custom pet is actually owned by the user. Ownership is only enforced in `console/user_pets_api.py`'s `UserPetListAPIView.patch` via `_selectable_pet_exists()` before it calls `UserPreference.update_known_preferences(...)`.

However, `KEY_USER_PET_SELECTED_ID` is registered as a normal entry in `UserPreference.PREFERENCE_DEFINITIONS`, so it's also reachable through the pre-existing generic endpoint `UserPreferencesAPIView.patch` (`console/api_views.py`), which calls `update_known_preferences` directly with no pet-specific ownership check. A user could PATCH `console/api/user/preferences/` with `{"preferences": {"user.pet.selected_id": "<another user's custom pet UUID>"}}` and it would normalize and persist successfully.

This is currently harmless in practice because every read path (`_resolved_pet_preferences` in `console/user_pets_api.py`) re-validates via `_selectable_pet_exists` and falls back to the default builtin pet when the stored id isn't owned — but that safety net lives in a different module than the write path, so it silently depends on every future consumer of this preference re-checking ownership rather than trusting the stored value. Consider enforcing ownership at write time (e.g. thread `user` through to the normalizer, since `update_known_preferences` already has it available) so an unowned/invalid pet id can never be persisted in the first place.

---

frontend/src/components/pets/PetSprite.tsx
```diff
@@ -0,0 +1,33 @@
...
+  spritesheetUrl,
+  row,
+  column,
+  className,
+  label,
+}: PetSpriteProps) {
+  const style = {
+    '--pet-spritesheet-url': `url("${spritesheetUrl.replaceAll('"', '%22')}")`,
+    '--pet-frame-x': `${(column / 7) * 100}%`,
+    '--pet-frame-y': `${(row / 10) * 100}%`,
```
matt-greathouse: Minor: `7` and `10` here (columns-1/rows-1) duplicate `PET_COLUMNS`/`PET_ROWS` from `api/services/user_pets.py`, and `ImmersivePetLayer.tsx` separately hardcodes `% 8` / `row < 8 ? 9 : 10` for the same grid. There's no single source of truth for the spritesheet grid dimensions between backend and frontend, so a future change to the grid size (e.g. `PET_COLUMNS`/`PET_ROWS`) would silently desync the frontend rendering unless someone remembers to update all three spots. Consider defining the grid dimensions once in `petAnimation.ts` (or a shared constants file) and importing them everywhere they're needed.

---

matt-greathouse (commented):
Reviewed the diff (backend models/migration/API, complexity-budget tooling, and the frontend pet components/hooks/integration points). Overall the feature is well-tested (private-file access, quota enforcement, spritesheet frame validation, preference normalization, position bounds) and the `ChatSidebar`/`SidebarSettingsMenu` context-menu refactor into a shared `FixedContextMenu` is a clean dedup.

Left 3 comments:
- A process concern: this PR introduces a `complexity-budget: exclude-*` self-exemption mechanism and immediately applies it to nearly all of its own new code, without the PR description explaining why.
- A defense-in-depth gap: pet-selector ownership is validated only in the dedicated pets endpoint, not in the shared preference normalizer, so the generic `/console/api/user/preferences/` endpoint can persist an unowned pet id (currently masked by a read-time fallback).
- A minor maintainability nit: spritesheet grid dimensions are hardcoded in three separate places across backend and frontend.

No correctness-blocking bugs found in the upload/validation/storage/deletion paths I reviewed.

---

PR #1389, opened 2026-07-25
Keep the composer auto-focus off touch-primary devices (#310)

ai-christianson:
## The report

Switching agents on mobile raises the on-screen keyboard unasked. Reported as a *"super important UX issue."*

## Root cause

`AgentComposer.tsx` auto-focuses the composer on every `agentId`/`focusKey` change:

```js
useEffect(() => {
  if (!autoFocus) return
  const timer = setTimeout(() => {
    textareaRef.current?.focus()   // this is what raises the keyboard
    ...
  }, 100)
}, [agentId, autoFocus, focusKey, moveTextareaCursorToEnd])
```

`AgentChatPage.tsx:4433` passes `autoFocusComposer` unconditionally, so it fires on every switch with no device check. Focusing a textarea *is* the API that opens the keyboard on iOS and Android.

Worth noting a full page **reload** does not reproduce it — only client-side switching does, which is why it shows up while using the app rather than on load.

## The fix: keep it where it helps, drop it where it hurts

Auto-focus now requires a **fine pointer**. Desktop still lands in the composer on switch; a phone or tablet leaves focus alone until the user taps the field. That matches what Slack, iMessage, WhatsApp and Telegram all do.

Deliberately `(hover: hover) and (pointer: fine)` rather than the `isTouchDevice` check already in this file — that one keys off touch *support*, so it would wrongly disable auto-focus on a touchscreen laptop.

It also no longer pulls focus out of another field already in use, which is the same principle applied to desktop.

## Verified in a real browser, not just jsdom

Local stack, iPhone 13 viewport, logged in, switching agents through the actual in-app switcher:

| Device | Pointer | Switch happened | `document.activeElement` after |
|---|---|---|---|
| Mobile, **before fix** | coarse | yes | **`textarea` in `.composer-shell`** → keyboard pops |
| Mobile, **after fix** | coarse | yes | **`body`** → no keyboard |
| Desktop, after fix | fine | yes | **`textarea` in `.composer-shell`** → convenience kept |

## Tests

Three behaviours, and **2 of the 3 fail without the fix** (verified by stashing it):

1. Touch-primary device: switching agents does not focus the composer
2. Fine pointer: switching agents still focuses it — guards against over-correcting
3. Focus is never pulled out of another field in use

Frontend suite **242/242**, `tsc --noEmit` clean. The 2 eslint warnings in this file are unchanged from `main`.

### DIFF
--- a/frontend/src/components/agentChat/AgentComposer.test.tsx
+++ b/frontend/src/components/agentChat/AgentComposer.test.tsx
@@ -328,6 +328,81 @@ describe('AgentComposer pending action insights panel', () => {
     expect(screen.queryByText('Which account should I use?')).not.toBeInTheDocument()
   })
 
+  function setPointer(kind: 'fine' | 'coarse') {
+    Object.defineProperty(window, 'matchMedia', {
+      configurable: true,
+      writable: true,
+      value: (query: string) => ({
+        matches: kind === 'fine' && query.includes('pointer: fine'),
+        media: query,
+        onchange: null,
+        addListener: () => {},
+        removeListener: () => {},
+        addEventListener: () => {},
+        removeEventListener: () => {},
+        dispatchEvent: () => false,
+      }),
+    })
+  }
+
+  function switchAgent(rerender: (ui: React.ReactElement) => void, focusKey: string) {
+    rerender(
+      <AgentComposer
+        onSubmit={vi.fn(async () => undefined)}
+        pendingActionRequests={[]}
+        insightsLoading={false}
+        isProcessing={false}
+        autoFocus
+        focusKey={focusKey}
+      />,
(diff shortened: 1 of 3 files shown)


ai-christianson:
**Complexity budget:** the first push failed `Check source LoC budget` at **+23**. I tightened the helper and the focus guard down to **+11**, which is the irreducible cost of the pointer check plus not-stealing-focus, then recorded the increase through the documented `--update-baselines` path.

Flagging it rather than letting it ride in the diff: `scripts/budgets/complexity_budgets.json` is bumped in this PR.

Tests and typecheck re-verified after the tightening — 40/40 on the composer suite, `tsc --noEmit` clean.

PR #1399, opened 2026-07-25
Retire mandatory agent Planning Mode

ai-christianson:
## Why

Planning Mode became a second runtime state machine around normal agent behavior. In the production trajectory that triggered this work, the model recognized that replacing the charter was risky, but `end_planning` required it; the accepted plan then overwrote durable role instructions. The same mechanism added special prompts, tool filtering, question rules, timeout recovery, and signup-preview gates, while active sessions could remain stuck.

## What changed

- New agents start in normal execution; legacy `planning` creation requests are normalized to `skipped`.
- Existing active planning rows migrate to `skipped`, and their stale pending planning questions are cancelled.
- Removed the planning-only prompt block, tool restrictions, schedule/config gates, timeout task, and event-loop corrections.
- Kept the legacy completion/skip API path as a safe compatibility shim: stale `end_planning` calls can finish, but can never replace the charter.
- Preserved user-requested planning as ordinary guided conversation. Explicit requests to shape a workflow may ask a bounded set of material questions, then wait.
- Simplified signup-preview gating so the upgrade panel appears only after the first reply is actually complete.
- Replaced the obsolete planning prompt complexity benchmark with the real first-run path and lowered the core source ceiling by 401 lines.

## Verification

- DeepSeek V4 Flash `guided_planning_micro`: **18/18** tasks across 3 runs per scenario (`609effdf-5023-440a-bfff-4d2be4b9a1b6`). This was rerun after removing the last added planning prompt sentence.
- DeepSeek V4 Flash full `all` suite: **1,575/1,621 (97.2%)** on the first stochastic pass (`44dc54c5-2b1e-484f-a36c-0b498f0a5070`). All 32 scenarios with a miss were replayed; 20 cleared completely and the sole provider error cleared 9/9. The remaining strict misses were compared against an isolated unchanged-main worktree: eight reproduced at the same task-level rate, and the branch-sensitive cases were separately replayed. No planning-specific regression reproduced.
- Focused backend batches: **884 tests green** (`batch_agent_chat`, `batch_event_processing_credits`, `batch_eval_fingerprint`, `batch_api_agents`, `batch_agent_webhooks`, `batch_schedule`).
- Frontend production build and touched-file ESLint: green.
- Django system check, compileall, migration check, diff check, and complexity budgets: green.

## Scope / compatibility

This intentionally does not remove persisted planning fields or every legacy UI renderer in one migration. Those remain inert for API/history compatibility. The behavioral state machine is retired; the migration and compatibility handler make old rows safe. Net diff is **968 fewer lines** overall and one commit.


ai-christianson:
Final preview verification on exact head `698e4ac4`:

- `https://pr-1399.ship.gobii.ai/api/v1/ping/` authenticated successfully.
- API-created contained QA agent persisted with `planning_state: skipped`.
- Authenticated headed Chrome/CDP check reached the agent page, cleared the one-time ticket fragment, rendered the probe and chat UI, and showed no login form or mandatory Planning Mode gate. The 1440×1000 screenshot was visually inspected.
- QA probe `be1a4c62-20f7-4182-aa00-5a90da7e93e4` was archived after validation.
- All GitHub checks are green and the branch contains current `main`.

PR #1404, opened 2026-07-25
Repair literal newline runs before sending to Discord (#228)

ai-christianson:
Fixes Troy's **#228**, reported by Andrew with a screenshot. Andrew noted he hadn't seen it recently and asked whether it was still real, so I checked production before writing anything.

## It is still real, but rarer than the ticket implies

Outbound Discord messages containing a literal backslash-n, from the analytics replica:

| day | with literal `\n` | total outbound discord |
|---|---|---|
| 2026-07-25 | 5 | 352 |
| 2026-07-24 | 1 | 302 |
| 2026-07-23 | 2 | 141 |
| 2026-07-21 | 2 | 78 |

So roughly **1–2% of messages**, ongoing, including today — not "every Discord user every time". That matches Andrew's sense that it had gone quiet. Worth fixing, not worth calling MAJOR.

Sampling the real bodies separates two classes:

- **Genuine defects** — `**#314 logged**\n\n**Status:** minor\n\nAndrew reports…`
- **Legitimate mentions** — bug reports that *quote* a literal `\n` in backticks while describing this very defect

Any fix has to repair the first and leave the second alone.

## Two defects, not one

**1. The #176 repair never ran on Discord.** `repair_structural_literal_newlines` was only ever applied to charters. Outbound message bodies were never covered.

**2. The repair was incomplete.** Its lookahead required the *next* token to be structural, so on the most common real case it repaired the first of a pair and stranded the second:

```
'**#314 logged**\n\\n**Status:** minor\n\\nAndrew reports...'   ← before: still broken
'**#314 logged**\n\n**Status:** minor\n\nAndrew reports...'     ← after
```

That is why this reads as a regression of a fix that shipped months ago: it never fully worked. A run of two or more is now treated as the paragraph break it can only be.

## Verification

Validated against the **actual production strings** pulled from the replica:

| case | before | after |
|---|---|---|
| `#314` paragraph breaks | 2 repaired, **2 still literal** | 4 repaired, **0 remaining** |
| `#288` paragraph break | 1 repaired, 1 remaining | 2 repaired, 0 remaining |
| bullet list | 2 repaired, 0 remaining | unchanged |
| backticked mention | left alone | **left alone** |
| prose mention | left alone | **left alone** |

- **8 new tests**, 2 red before the change.
- Charter tests, Discord tests, and the 347 tests in modules consuming this helper all still pass — the helper is shared with charter repair, so this improves that path too.

Second commit records the +25 LoC budget.

Preview validation to follow before merge.

### DIFF
--- a/api/agent/tools/charter_text.py
+++ b/api/agent/tools/charter_text.py
@@ -2,6 +2,14 @@ import re
 
 
 LITERAL_NEWLINE = "\\n"
+# Two or more in a row is a paragraph break. Nothing writes that on purpose, and matching the whole
+# run matters: a lookahead that only inspects the next token repairs the first of a pair and leaves
+# the second stranded, which is how "\n\n## Heading" used to come out as a real break followed by a
+# visible \n.
+LITERAL_NEWLINE_RUN_RE = re.compile(r"(?:\\n){2,}")
+# A single one is only unambiguous when what follows can only begin a line: another break, a
+# heading, a bullet, or a numbered item. A lone \n in prose -- including one inside backticks being
+# discussed as text -- is left exactly as written.
 STRUCTURAL_LITERAL_NEWLINE_RE = re.compile(
     r"\\n(?=(?:\\n|#{1,6}(?:\s|$)|[-*+]\s|\d+[.)]\s))"
 )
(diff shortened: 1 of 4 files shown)


ai-christianson:
**Preview validation: partially possible, and I want to be explicit about what I did not do.**

The preview came up and authenticates. What I could not do is exercise the change end to end, because the only path through it is actually sending a Discord message — and a preview is a fork carrying **production Discord credentials**, so that would post to a real server. Sending spam into a live channel to test a formatting fix is the wrong trade, so I did not.

Nor is the deployed commit externally identifiable: there is no version or build endpoint, and nothing user-visible changes except Discord output itself.

What that leaves, and why I judged it sufficient here:

- This is a **pure, deterministic string transformation**. It takes text and returns text, with no dependence on infrastructure, timing, context or scale. That is categorically unlike #1398, where the defect only appeared against a production-shaped fork and preview validation was essential.
- It is verified against the **actual production strings** pulled from the analytics replica, not invented fixtures — including the two classes that must behave differently (genuine paragraph breaks vs. bug reports quoting a literal `\n` in backticks).
- 8 unit tests, 2 red before the change; charter, Discord and consumer-module suites all still green; 17/17 CI.

If someone wants belt and braces, the safe follow-up is to watch the same replica query after this ships: outbound Discord messages containing a literal backslash-n should trend to only the legitimate mentions.

PR #1405, opened 2026-07-25
Fix the newline contract and remove the output rewriting (#228)

ai-christianson:
Andrew asked whether #1404 was a layered-on hack or a root cause fix. **It was a hack**, and a partial one. This removes it and fixes the cause instead.

## What #1404 did wrong

It rewrote model output before sending — the platform-side output filtering `AGENTS.md` explicitly rules out. I found `charter_text.py` already doing this for charters and treated that precedent as permission. A precedent is not a justification; that helper may itself have been the original mistake.

It was also only partial. Prose line breaks still shipped literally:

```
in : 'Hi Ben,\n\nQuick question about territory.\n\nBest,\nBelle'
out: 'Hi Ben,\n\nQuick question about territory.\n\nBest,\\nBelle'   ← signature still literal
in : 'Address:\n123 Main St\nBoston MA'
out: unchanged — both still literal
```

A real outbound email still rendered `Best,\nBelle` to a prospect. It looked fixed, which is worse than untouched.

## What this does

**Removes the runtime rewriting entirely.** Not kept as a mitigation — stacking a contract fix on top of an output filter is two layers doing the same job, and the wrong one silently wins.

**Fixes the actual cause.** The `message` contract described markdown sections, bullets, bold labels, tables, HTML and XML, and said nothing about newlines. It even warned that XML "is sent literally", so the concept was present. It demanded structure that requires line breaks without ever saying how to express one. That is an information gap.

**Adds eval coverage.** The message-quality suite now deterministically fails a body that uses literal `\n` where a line break was meant, reading the **model's own tool parameters** — so it measures behavior, and there is no repair layer left to mask it.

The helper survives only in the explicit `repair_agent_charter_newlines` management command — a deliberate one-off tool, not silent runtime rewriting — and the run-handling fix makes that command work correctly on paragraph breaks it previously half-repaired.

## What I can prove, and what I cannot

Analytics replica, 30 days of outbound Discord:

| group | n | avg length | avg bold markers |
|---|---|---|---|
| clean | 1434 | 599 | 6 |
| contains literal `\n` | 24 | 768 | 12 |

Base rate **1.6%**, correlated with markdown density — offending bodies carry twice the bold markers and run 28% longer.

**No feasible number of eval runs can resolve that.** Detecting 1.6% → 0.4% at 80% power needs ~1,500 runs per arm; even 1.6% → 0% needs ~250. The standing bar is ≥16 per arm, which measures nothing here.

So, per `AGENTS.md` — *"if measurement shows no effect, land the eval as coverage and leave the bug open"*:

- the contract change ships as a **justified information fix**, defensible on inspection
- the eval ships as **coverage** that will catch any regression raising the rate
- it is **not** reported as a proven fix, and **#228 stays open**

With the rewriting removed, the defect is now visible again when it occurs, which is the point — a symptom the platform hides cannot be measured or fixed.

## Verification

- 60 tests across the discord, charter and newline modules: green.
- Eval harness smoke-run on `message_quality_chat_project_update`, DeepSeek V4 Flash: 8/8 tasks passed including the new assertion, confirming no false positives on clean output.
- Complexity budget recorded.

### DIFF
--- a/api/agent/tools/charter_text.py
+++ b/api/agent/tools/charter_text.py
@@ -29,3 +29,14 @@ def repair_structural_literal_newlines(value: str | None) -> tuple[str, int, int
     repaired = STRUCTURAL_LITERAL_NEWLINE_RE.sub("\n", repaired)
     remaining = count_literal_newlines(repaired)
     return repaired, before - remaining, remaining
+
+
+def literal_newline_failure(value: str | None) -> bool:
+    """True when a body uses literal backslash-n where a line break was meant.
+
+    Only the unambiguous positions count: a run of them, or one before a heading, bullet or
+    numbered item. A lone occurrence in prose -- someone writing about the escape sequence -- is
+    not a formatting failure.
+    """
+    text = value or ""
+    return bool(LITERAL_NEWLINE_RUN_RE.search(text) or STRUCTURAL_LITERAL_NEWLINE_RE.search(text))
--- a/api/agent/tools/send_discord_message.py
+++ b/api/agent/tools/send_discord_message.py
@@ -42,7 +42,11 @@ def get_send_discord_message_tool() -> Dict[str, Any]:
                                        "Discord cannot render tables: never send pipe-separated columns with a "
                                        "hyphen-divider row, even as a summary. "
                                        "Use Markdown only; raw HTML is rejected. Use code formatting to show HTML literally. "
-                                       "Do not pass tool-call/XML syntax; it is sent literally.",
+                                       "Do not pass tool-call/XML syntax; it is sent literally. "
+                                       "Write line breaks as real newlines in the string value. A backslash "
+                                       "followed by n is sent literally and the reader sees those two "
+                                       "characters instead of a line break, which breaks every heading, "
+                                       "bullet and paragraph above.",
                     },
                     "attachments": {
                         "type": "array",
--- a/api/agent/tools/send_discord_message.py
+++ b/api/agent/tools/send_discord_message.py
@@ -10,7 +10,6 @@ from api.agent.files.attachment_helpers import AttachmentResolutionError, resolv
 from api.agent.comms.outbound_content_policy import markdown_only_error
 from api.agent.tools.attachment_guidance import SEND_TOOL_ATTACHMENTS_DESCRIPTION
 from api.agent.tools.agent_variables import substitute_variables_with_filespace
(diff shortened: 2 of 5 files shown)


ai-christianson:
**Preview validated** on `pr-1405.ship.gobii.ai`: authenticates as andrew@gobii.ai, MCP surface healthy (15 tools).

As with #1404, the one thing a preview cannot safely exercise here is the outbound Discord send itself — a fork carries **production Discord credentials**, so driving that path posts to a real server. I am not sending test traffic into a live channel to check a formatting contract.

What that leaves is sufficient for this change, and for a specific reason: **there is no longer any runtime code to validate.** The output rewriting is deleted (`grep` on the send path returns 0). What ships is a tool-parameter description the model reads, plus an eval assertion — both verified by the harness rather than by delivery:

- eval harness smoke-run on `message_quality_chat_project_update` with DeepSeek V4 Flash: **8/8 tasks passed**, including the new `verify_formatting_basics` assertion, confirming it does not false-positive on clean output
- 60 tests across the discord, charter and newline modules green
- 16/16 CI

The behavioural claim is deliberately empty: at a **1.6% base rate** nothing here is proven, #228 stays open, and with the rewriting removed the defect is observable again — which is the point.

ai-christianson:
**Correcting the evidence in this PR.** The body claimed verification on the strength of a single 2-run smoke test. That was thin, and step 8 of the eval workflow requires the affected suite. I have now run it.

**Suite `message_quality_reports`, DeepSeek V4 Flash, run `ee030d48-1b4c-4998-b447-32acf4b617d7`: 72/73 tasks, 98.6%.**

The one failure is not the new assertion:

    name    : judge_message_quality
    observed: LLM judge: Fail. "The email is essentially raw HTML with bold tags and line
              breaks, but it lacks any real visual polish. No tables, metric blocks, inline
              color accents, emoji/icons, badges, or status markers."

That is the subjective richness judge on an email scenario — stochastic, and unrelated to newline handling.

The signal that matters: **`verify_formatting_basics` passed on every scenario in the suite**, so the new `literal_newline_failure` check produces no false positives across bodies full of genuine markdown structure — headings, bullets, bold labels, multi-paragraph reports. That was the real risk of adding it, and it is clear.

Still unchanged: this proves the assertion is safe, **not** that the contract change fixes anything. At a 1.6% base rate that remains unmeasurable, and #228 stays open.

PR #1406, opened 2026-07-25
Key the agent roster cache by the agent it was scoped to (#363)

ai-christianson:
Fixes **#363**, which I found and filed while validating #1398 on a preview but did not fix at the time.

## The symptom

Open a link to an agent in an organization while your active context is personal, and the sidebar shows **one agent** — that agent — while the server returned the full roster. Nothing indicates a problem: `rosterLoading` stays pinned true forever.

Network trace from preview `pr-1398`:

```
1797ms  200  /console/api/agents/roster/                       ctx: personal,      agents: 0
4008ms  200  /console/switch-context/?for_agent=957e0a0b...    ctx: organization
```

An earlier trace of the same path showed the scoped roster returning **105 agents** — all discarded.

## Root cause

`useAgentRoster` passes three inputs to the fetch but keys the cache on two:

```ts
queryKey: ['agent-roster', contextKey],                          // context + staffContext only
queryFn: () => fetchAgentRoster({ context, forAgentId, staffContext }),
```

`context` and `staffContext` are folded into `contextKey`. **`forAgentId` is not.**

A roster scoped to an agent comes back carrying *that agent's own console context*, which need not match the context the key describes. So the scoped response was filed under the unscoped entry, and with `placeholderData: keepPreviousData` it was served back for a key it did not belong to. The caller then compared `rosterQuery.data.context` against `effectiveContext`, saw a mismatch, and emptied the list:

```ts
const agents = contextReady && !rosterContextMismatch ? rosterQuery.data?.agents ?? [] : []
```

This is a cache-key correctness bug, not a context-switching bug: the key must capture every input the query function depends on.

## Fix

Include `forAgentId` in the query key. One line, plus the comment explaining why.

Invalidation is unaffected — every caller uses the bare `['agent-roster']` prefix without `exact: true`, and prefix matching still covers the widened key. There is a test asserting exactly that, because it was the obvious way for this change to break something.

## Verification

**4 tests, 2 red on `main`** (verified by stashing the hook and re-running):

- a different route agent triggers its own fetch rather than reusing a foreign entry
- an unscoped caller is not served an agent-scoped roster
- an unchanged context and agent still reuses the cache — the fix must not defeat caching
- the entry is still reachable by the bare prefix callers invalidate with

I initially wrote these as key-construction assertions, which would have passed on `main`. Rewrote them to drive the real hook.

- Frontend suite **268/268**, `tsc -b --force` clean, lint identical to `main` (201 both).

## Note on scope

This addresses the cache collision that produced the observed failure. It does **not** rework how `rosterContextMismatch` handles a genuinely stale context — if that state is still reachable another way, the list will still empty, and that remains worth its own look. Preview validation below is what decides whether the observed symptom is gone.

### DIFF
--- a/frontend/src/hooks/useAgentRoster.ts
+++ b/frontend/src/hooks/useAgentRoster.ts
@@ -18,11 +18,16 @@ export function useAgentRoster(options?: UseAgentRosterOptions) {
   const contextKey = options?.contextKey ?? 'default'
   const refetchIntervalMs = options?.refetchIntervalMs ?? false
 
+  // The roster scoped to a specific agent comes back with that agent's own context, which can
+  // differ from the one the key describes. Leaving forAgentId out of the key filed that response
+  // under the wrong entry, and the caller then discarded it as a context mismatch.
+  const forAgentId = options?.forAgentId ?? null
+
   return useQuery({
-    queryKey: ['agent-roster', contextKey] as const,
+    queryKey: ['agent-roster', contextKey, forAgentId] as const,
     queryFn: () => fetchAgentRoster({
       context: context ?? undefined,
-      forAgentId: options?.forAgentId,
+      forAgentId: forAgentId ?? undefined,
       staffContext: options?.staffContext,
     }),
     placeholderData: keepPreviousData,
--- a/frontend/src/hooks/useAgentRosterScoping.test.tsx
+++ b/frontend/src/hooks/useAgentRosterScoping.test.tsx
@@ -1,0 +1,91 @@
+/**
+ * Regression coverage for #363: the agent roster was fetched and then discarded.
+ *
+ * A roster scoped to a specific agent (`for_agent`) comes back carrying that agent's own console
+ * context, which can differ from the context the cache key describes. `forAgentId` was missing from
+ * the query key, so the scoped response was filed under the unscoped entry and served back for a
+ * key it did not belong to. The caller compared the two contexts, called it a mismatch, and emptied
+ * the list — a sidebar showing one agent when the server had returned a hundred.
+ */
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { renderHook, waitFor } from '@testing-library/react'
+import type { PropsWithChildren, ReactElement } from 'react'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { ConsoleContext } from '../api/context'
+import { useAgentRoster } from './useAgentRoster'
+
+const { fetchAgentRosterMock } = vi.hoisted(() => ({ fetchAgentRosterMock: vi.fn() }))
+
+vi.mock('../api/agents', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../api/agents')>()),
+  fetchAgentRoster: fetchAgentRosterMock,
+}))
+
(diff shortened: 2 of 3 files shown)


ai-christianson:
**Preview validation failed. This does not fix #363 and I am not merging it as one.**

Validated on `pr-1406.ship.gobii.ai` with an authenticated session, driving the exact path that produced the bug — opening an organization agent by URL from a personal context:

    2079ms 200 /console/api/agents/roster/   agents=0  ctx=personal

    t=3s    rosterRows: 1, countPending: 1
    t=9s    rosterRows: 1, countPending: 1
    t=15s   rosterRows: 1, countPending: 1

Still one agent, still pinned loading, fifteen seconds in. Unchanged from the broken behaviour.

**What that tells me.** In this run the agent-scoped roster request never fired at all — only the unscoped one. So the list is not empty because a scoped response landed in the wrong cache entry; it is empty because the scoped query is gated behind `enabled: contextReady`, and `contextReady` never becomes true on this path. My cache-key change cannot help when the query never runs.

**The cache-key bug is still real** — `forAgentId` is passed to the query function but was absent from the query key, and 2 of the 4 tests here fail on `main`. It is a genuine correctness defect and worth fixing. It is simply **not the cause of #363**, and I was wrong to assume it was.

Options, and I would rather be told than choose silently:

1. Retitle and merge this as the narrow cache-key correctness fix it actually is, explicitly not claiming #363.
2. Close it and fold the change into whatever properly fixes `contextReady`.

#363 stays open either way. The next person needs to start at why `contextReady` stays false when a route agent belongs to a different console context — that is the real defect, and it is upstream of everything I touched here.

ai-christianson:
**Correction: my previous comment was wrong. This does fix #363, and I called failure too early.**

I probed at a 15-second cutoff during a run where the scoped roster query took longer than that to return, saw one row, and concluded the fix did nothing. Same mistake as earlier tonight — concluding from insufficient observation rather than from the sequence.

Full request sequence on `pr-1406.ship.gobii.ai`, authenticated, opening an organization agent by URL from a personal context:

     4403ms  REQ   GET /console/api/agents/roster/?for_agent=957e0a0b...
     4403ms  RESP  200 /console/switch-context/?for_agent=...   ctx=organization
    10960ms  RESP  200 /console/api/agents/roster/?for_agent=... agents=116 ctx=organization
    11901ms  ROSTER rows=116  pending=0  top=116

**The scoped request now fires at all**, which is the fix. Before, `forAgentId` was absent from the query key, so changing the route agent did not change the key and React Query never refetched — the scoped roster was simply never requested, and the sidebar kept the unscoped personal result of zero agents plus the one agent from the URL.

**Repeated three times, identical:**

| run | roster response | rendered |
|---|---|---|
| 1 | 11424ms, 116 agents | rows=116, pending=0, top=116 |
| 2 | 10613ms, 116 agents | rows=116, pending=0, top=116 |
| 3 | 10603ms, 116 agents | rows=116, pending=0, top=116 |

## A second, separate finding

**That query takes 6.5 seconds server-side** for 116 agents, consistently. The sidebar therefore sits on a loading placeholder for ~11 seconds before the roster appears. That is not a regression from this change and it is not what #363 described, but it is a real performance problem that only shows at production cardinality — a local seed of two agents returns instantly. Worth its own ticket; I am not folding it in here.

Merging on the strength of the three runs above.

PR #1407, opened 2026-07-26
Cut the agent roster's per-agent queries (partial)

ai-christianson:
Found while validating #1406 on a preview: opening the agent chat left the sidebar on a loading placeholder for **~11 seconds** for an organization with 116 agents. A local seed of two agents returns instantly, which is why this never showed up before.

## Measured, on the analytics replica, against a real organization

The roster's latest-message lookup alone:

```
Bitmap Heap Scan on api_persistentagent   (actual time=24.401..8153.569 rows=144)
  ->  Limit                (actual time=55.806..55.806 loops=144)   ← once per agent
      ->  Sort             (actual time=55.803..55.803 loops=144)
          ->  Bitmap Heap Scan on ...message (rows=444 → 99, loops=144)
                Index: api_persistentagentmessage_owner_agent_id_fae990f3
Execution Time: 8154.282 ms
```

**8.15 seconds**, and the code paid it **twice** — `id` and `timestamp` were separate `Subquery()` expressions, so two correlated subqueries per agent.

## Why the index did not help

`pa_msg_agent_ts_idx (owner_agent_id, timestamp DESC)` exists and looks perfect for this. The planner cannot use it, because the visibility predicates are:

- `raw_payload__hide_in_chat` — a **JSONB** test, not indexable here, requiring heap access
- `NOT (peer_agent_id IS NOT NULL OR conversation.is_peer_dm)` — a **join to conversations** inside an `exclude()` of an OR

So each agent degraded to a bitmap heap scan of ~444 of its messages, filtered to 99, sorted, `LIMIT 1`. 55.8ms × 144.

## The change

One ordered pass over the candidate rows, taking the first row seen per agent. Same data, same replica:

| | time |
|---|---|
| correlated subquery per agent | **8154 ms** |
| single ordered pass | **897 ms** |

**~9× faster**, and the real saving is larger because the old form ran twice.

`DISTINCT ON (owner_agent_id)` expresses this directly and is what I wrote first — but it is **PostgreSQL only**, and both `config/test_settings` and `config/eval_local_settings` are SQLite, so it would have broken every test and the eval environment. The dedupe streams in Python instead.

## Tests

The helper had **no direct coverage** despite carrying non-obvious exclusions — hidden messages, peer DMs — and it drives what every roster row displays.

Added seven tests pinning the contract: latest outbound wins, inbound ignored, hidden ignored, peer DMs ignored, per-agent isolation, absent-message case, and an entry returned for every requested agent.

**They pass against both the old and the new implementation.** That is deliberate: this is a refactor, so the tests prove equivalence rather than red-to-green. Running them against the old code also caught two bugs in my own fixtures before they could mask a real difference.

Adjacent suites green: `test_agent_chat_api`, `test_agent_chat_access`, `test_agent_chat_signals`.

## Scope

This fixes the dominant cost. The roster response also does other per-request work, and the ~11s user-visible wait includes more than this query — I have not claimed the whole wait disappears, only that the largest single contributor drops from 8.15s to 0.9s. Preview timing to follow before merge.


ai-christianson:
**Update: my original diagnosis was incomplete, and the first fix alone did not move the endpoint.**

I measured a single query in isolation on the replica (8154ms → 897ms) and assumed it was the bottleneck. Then I measured the **endpoint** on the preview and it still took 13.6s. I had optimised the thing I measured rather than the thing that was slow.

Measuring query *counts* found the real problem: the roster issued roughly **1.5 queries per agent** on top of a fixed cost — about 200 round trips at 116 agents.

**Two per-agent queries, both fixed at source:**

1. `preferred_llm_tier` is a foreign key that was not in `select_related`, so building each payload fetched an `IntelligenceTier` row. The `select_related` existed on one branch and was **lost on the branch that actually runs**, where the queryset is rebuilt from `agent_queryset_for`.
2. `is_personal_trial_enforcement_enabled()` queried the waffle switch table on **every call** — a single global boolean, resolved once per agent. Waffle caches its switches by default; this lookup queried the model directly and so opted out of that cache. Now cached briefly, **including the absent-switch case**, which is the common state and is why my first attempt at this changed nothing.

**Measured locally, warm:**

| agents | before | after |
|---|---|---|
| 3 | 30 queries | **26** |
| 5 | 33 queries | **27** |

Per-agent slope **1.5 → 0.5**, and the residual is one bounded `messageread` lookup rather than per-agent work. Projected at 116 agents: ~200 queries → ~27.

**On the earlier `message_reads` change in this PR:** still a genuine improvement (8154ms → 897ms for that query on real data, and the old form ran twice), with seven tests passing against both implementations. But it was **not** the dominant cost and I should not have implied it was.

Tests: 4 new for the switch cache, 2 red without it. `test_agent_chat_api`, `test_agent_chat_access`, `test_agent_chat_signals`, `test_agent_transfer`, `test_human_input_requests`, `test_agent_collaborators`, `test_pages` and the trial suites all green. The cache cannot affect tests using `@override_settings(PERSONAL_FREE_TRIAL_ENFORCEMENT_ENABLED=True)` — the env override short-circuits ahead of it, verified by test.

Preview timing to follow before merge.

ai-christianson:
**Final scope, and a correction to two things I claimed earlier in this PR.**

**Correction 1 — I broke tests and misdiagnosed it as pre-existing.** I tried caching the trial-enforcement switch, first by hand and then via waffle's `switch_is_active`. Both break tests that create a switch and expect it observed immediately. I convinced myself the failures were pre-existing by running `git stash push` on files I had **already committed** — so it stashed nothing and I compared my branch against itself, twice. Checking the files out from `origin/main` instead: both tests **pass on main and fail on my branch**. Mine. Backed out.

**Correction 2 — the roster query I opened this PR with was not the dominant cost.** It is a real improvement (8154ms → 897ms on production-scale data, and the old form ran twice), but the endpoint stayed at 13.6s until I measured query *counts* rather than one query in isolation.

## What this PR now contains

1. **`message_reads`**: one ordered pass instead of a correlated subquery per agent, run twice. 8154ms → 897ms measured on the replica. Seven new tests, passing against both implementations, since it is a refactor.
2. **`select_related("preferred_llm_tier")`** on the roster queryset, including the branch that actually runs — the existing `select_related` was lost when the queryset is rebuilt from `agent_queryset_for`. Removes one `IntelligenceTier` fetch per agent.

## What remains, measured and unfixed

`is_personal_trial_enforcement_enabled()` queries the waffle switch table **once per agent** — one global boolean, ~116 round trips for a large organization. Per-agent slope is **1.5 queries/agent**, of which this is 1.0.

**Caching is the wrong fix and I proved it twice.** The correct fix is to resolve the flag once per request and pass it down through `user_can_manage_agent` / `_is_blocked_personal_owner`, so no second cache exists whose invalidation nobody owns. That is a wider signature change than this PR should carry, and it deserves its own review.

So the ~11s user-visible wait is **improved, not solved**. I would rather land the measured, safe part and write the rest up accurately than claim the whole thing.

Also worth a ticket separately: `test_waffle_switch_skips_initial_free_plan_credit_grant` and `test_waffle_switch_enforcement_only_returns_grandfathered_free_users` both fail when run in isolation on `main` — they depend on test ordering. They pass in suite context, which is why CI is green.

ai-christianson:
**Preview validated. Roster request time roughly halves.**

Measured on `pr-1407.ship.gobii.ai` with an authenticated session against a real organization of 126 agents, timing the `roster/?for_agent=` request itself.

| | runs (ms) | median |
|---|---|---|
| baseline (pr-1406, no fix) | 10603, 10613, 11424 | **~10.6s** |
| this branch | 3031, 3198, 4323, 5222, 6385, 9501 | **~4.8s** |

One further run at 14571ms was the first request after a fresh deploy — cold caches and connection pool — and is excluded as unrepresentative. The remaining spread (3.0–9.5s) is wider than the baseline's, which was tight; the endpoint still does real work and contends with a shared preview.

**What landed, in the end:**

1. `message_reads`: one ordered pass instead of a correlated subquery per agent run twice. 8154ms → 897ms on the replica. Seven tests, passing against both implementations since it is a refactor.
2. `select_related("preferred_llm_tier")` on the branch that actually runs — the existing hint was lost where the queryset is rebuilt from `agent_queryset_for`.
3. Personal-agent access resolved **once per user instead of once per agent**, memoized on the user instance.

Per-agent query slope **1.5 → 0.5**, the residual being one bounded lookup rather than per-agent work.

**Three attempts at that third fix, and why only the last is safe.** A hand-rolled cache and waffle's own `switch_is_active` both broke tests that create a switch and expect it observed immediately — and the waffle form additionally made the browser-session-ticket endpoint return **500** on the preview, which I only caught because minting failed there. A memo on the user instance lives exactly as long as the request: no shared cache, nothing to invalidate, no cross-test bleed. Verified against `test_api`, `test_subscription_helper`, `test_agent_chat_access`, `test_agent_chat_api`, `test_agent_transfer`, `test_agent_collaborators`, `test_human_input_requests`, `test_pages`.

**Not claimed:** that the sidebar is now fast. ~4.8s median is better, not good. The endpoint does other work at this cardinality and deserves its own pass.

Also for a separate ticket: `test_waffle_switch_skips_initial_free_plan_credit_grant` and `test_waffle_switch_enforcement_only_returns_grandfathered_free_users` pass in suite context but fail in isolation on `main` — they depend on test ordering.

PR #1409, opened 2026-07-26
Meet the minimum touch target size on mobile chat controls (#337)

ai-christianson:
Fixes **#337**. WCAG 2.5.8 Level AA requires a 24×24 CSS px minimum for pointer targets. Verified still reproducing on current `main` before changing anything.

## Measured at 390px, before

| control | size |
|---|---|
| email header icon | **20 × 20** |
| message actions (copy / 👍 / 👎 / report) | **23.2 × 23.2**, four of them **4px** apart |
| insight tabs | 58 × **22** |
| agent name button | 76 × **16.3** |

The action cluster is the one that matters. Four controls at a **27px centre-to-centre pitch** against a fingertip closer to 40px — a mis-tap there doesn't miss, it **records the wrong feedback or opens a report**.

## After

| viewport | controls below minimum | action size | gaps | overflow |
|---|---|---|---|---|
| 390 × 844 | **none** | 28px | 8, 8, 8 | no |
| 375 × 667 | **none** | 28px | 8, 8, 8 | no |

Scoped to `(pointer: coarse)`, so precise pointers keep the denser layout — the same idiom already used in this file for pointer-aware behaviour.

The name button gains padding with a **matching negative margin**, so the target grows without moving anything around it. Confirmed by screenshot: the header is visually unchanged.

Verified by hit-testing rather than by reading `getBoundingClientRect`, which was necessary — my first attempt used a pseudo-element that measured "fixed" but was not actually clickable (17px real), and my selector targeted `.banner-name`, a `<span>` **inside** the button, rather than `.banner-name-button`. Both only surfaced because I probed the true clickable region with `elementFromPoint`.

## Deliberately not fixed

**Desktop still has controls under the minimum** — sidebar collapse 16×30, favourite 14×27, and the same chat controls at their denser sizes. Real, and I have the measurements, but this ticket was a mobile audit and closing the desktop gap changes visual density enough to deserve its own review. Worth a follow-up ticket rather than smuggling it in here.

## Testing

**No automated regression test.** jsdom performs no layout and does not evaluate media queries, so nothing in the suite can observe a target's rendered size. Everything above was measured in a real browser. A CSS-source assertion would only restate the rule and would not have caught either of the two mistakes described above. This is the gap tracked as #327.

- Frontend suite 268/268, `tsc -b --force` clean, lint identical to `main` (201 both).
- Stylesheet brace-balanced (checked explicitly after #1398).

### DIFF
--- a/frontend/src/components/agentChat/toolDetails/shared.tsx
+++ b/frontend/src/components/agentChat/toolDetails/shared.tsx
@@ -120,7 +120,9 @@ export function TruncatedMarkdown({ content, maxLines = 3 }: { content: string; 
       <button
         type="button"
         onClick={() => setIsExpanded(!isExpanded)}
-        className="inline-flex items-center gap-1 text-xs font-medium text-indigo-600 hover:text-indigo-700 transition-colors"
+        // py-1/-my-1 lifts the target to the 24px minimum without moving the card's layout: the
+        // text alone gives a 16px line box, which is under it.
+        className="inline-flex items-center gap-1 py-1 -my-1 text-xs font-medium text-indigo-600 hover:text-indigo-700 transition-colors"
       >
         {isExpanded ? (
           <>
--- a/frontend/src/styles/agentChatLegacy.css
+++ b/frontend/src/styles/agentChatLegacy.css
@@ -3888,6 +3888,36 @@
     color: #64748b;
     transition: border-color 0.15s ease, color 0.15s ease, background-color 0.15s ease;
 }
+/* WCAG 2.5.8 sets a 24x24 CSS px minimum for pointer targets. Several chat controls sit under it,
+   which matters most on a phone: the four message actions were 23.2px with 4px between them, so a
+   fingertip covered three at once and a mis-tap recorded the wrong feedback rather than missing.
+   Grow the hit area on coarse pointers; precise pointers keep the denser layout. */
+@media (pointer: coarse) {
+    .banner-contact-link {
+        width: 1.5rem;
+        height: 1.5rem;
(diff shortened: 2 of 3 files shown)


ai-christianson:
**Preview validated on real production-scale data. Zero controls below the minimum.**

Authenticated session on `pr-1409.ship.gobii.ai`, 390px viewport, two runs:

    run 1: belowMin=[]  actionSize=28  gaps=[8 x18]  overflow=false
    run 2: belowMin=[]  actionSize=28  gaps=[8 x18]  overflow=false

Eighteen action gaps, every one 8px — the fork has many message clusters, so this exercises far more of the surface than my local seed did.

**The preview earned its keep here.** The first validation run surfaced a control my local testing could not: `"Read full assignment 138x16"`, the assignment-card disclosure toggle. The fork has assignment cards and my seed did not, so it only appeared against real data. Fixed in the same way — padding with a matching negative margin — and the second round is clean.

Header screenshot confirms no visual shift from the padding changes.

Recap of what this closes, measured before and after at 390px:

| control | before | after |
|---|---|---|
| email header icon | 20 × 20 | ≥ 24 |
| message actions ×4 | 23.2, 4px apart | **28, 8px apart** |
| insight tabs | 58 × 22 | ≥ 24 |
| agent name button | 76 × 16.3 | 24.3 |
| assignment disclosure | 138 × 16 | ≥ 24 |

Still deliberately out of scope: desktop, which retains sidebar collapse at 16×30, favourite at 14×27 and the denser chat controls. Measured, real, and worth its own ticket rather than folded in here.

PR #1411, opened 2026-07-26
Bring desktop chat controls up to the minimum target size

ai-christianson:
Completes the accessibility gap I deliberately left open in **#337 / PR #1409**, which fixed mobile behind a `(pointer: coarse)` query. On a precise pointer, **11 controls remained under the WCAG 2.5.8 minimum of 24×24 CSS px**, measured identically at 1440px and 1280px.

## Two of them were a layout bug, not a design choice

| control | declares | measured |
|---|---|---|
| `chat-sidebar-toggle` | `1.9rem` square (30.4px) | **16 × 30** |
| `agent-roster-item__favorite` | square from its own variable | **14 × 27** |

Both were being **compressed by their flex parents**. `flex-shrink: 0` restores the size they already ask for — which is also what they visually ought to be. The sidebar toggles now render as proper squares instead of squeezed slivers.

## The rest now meet the minimum for every pointer

| control | before | after |
|---|---|---|
| contact icon | 20 × 20 | 24 × 24 |
| message actions ×4 | 23.2 | 24 |
| insight tabs | 58/67 × **22** | × 24 |
| name button | 82 × **18** | 24 (padding + matching negative margin, no layout shift) |

Mobile keeps its roomier 28px actions with 8px between them — that is a spacing/ergonomics choice, distinct from the conformance minimum, so the coarse-pointer block stays.

## Result

| viewport | controls below minimum |
|---|---|
| 1440 × 900 | **0** |
| 1280 × 800 | **0** |
| 390 × 844 | **0** |
| 375 × 667 | **0** |

Desktop layout confirmed unchanged by screenshot — the changes are 0.8–4px on icon boxes plus two elements no longer being squeezed.

- Frontend suite 268/268, `tsc -b --force` clean, lint identical to `main` (201 both).
- Stylesheet brace-balanced (checked explicitly, after #1398).

## Testing note

**No automated regression test.** jsdom performs no layout and cannot observe a rendered target size, so nothing in the suite can assert this. All numbers above are measured in a real browser. This is the gap tracked as #327.

### DIFF
--- a/frontend/src/styles/agentChatLegacy.css
+++ b/frontend/src/styles/agentChatLegacy.css
@@ -735,6 +735,12 @@
 }
 .banner-name-button {
     overflow: hidden;
+    /* The text line box alone is under the 24px minimum. Pad the target and take the same amount
+       back as negative margin, so it grows without moving the header around it. */
+    padding-top: 0.25rem;
+    padding-bottom: 0.25rem;
+    margin-top: -0.25rem;
+    margin-bottom: -0.25rem;
 }
 .banner-emotion {
     font-size: 0.875rem;
@@ -750,8 +756,9 @@
 }
 .banner-contact-link {
     display: inline-flex;
-    width: 1.25rem;
-    height: 1.25rem;
+    /* 24px is the minimum target size; the icon inside keeps its own dimensions. */
+    width: 1.5rem;
+    height: 1.5rem;
     align-items: center;
     justify-content: center;
     border-radius: 999px;
@@ -2499,9 +2506,9 @@
     align-items: center;
     justify-content: center;
     gap: 0.25rem;
-    height: 22px;
+    height: 24px;
     border: none;
-    border-radius: 11px;
+    border-radius: 12px;
     background: rgba(255, 255, 255, 0.54);
     cursor: pointer;
     padding: 0 10px;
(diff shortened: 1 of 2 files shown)


ai-christianson:
**Preview validated on real production data. Zero controls below the minimum at every viewport.**

Authenticated session on `pr-1411.ship.gobii.ai`, one fresh ticket per viewport (they are single-use — my first attempt reused one and the later contexts silently landed on the public marketing page, which is why I now assert `inApp` before trusting a result):

    1440x900  auth=/app/  {"inApp":true,"belowMin":[],"overflow":false}
    1280x800  auth=/app/  {"inApp":true,"belowMin":[],"overflow":false}
     390x844  auth=/app/  {"inApp":true,"belowMin":[],"overflow":false}

That covers both desktop widths and confirms the mobile work from #1409 did not regress.

Recap of what this closes:

| control | before | after |
|---|---|---|
| sidebar toggle | **16 × 30** (declared 1.9rem square, compressed by flex) | 30 × 30 |
| roster favourite | **14 × 27** (same compression) | square |
| contact icon | 20 × 20 | 24 |
| message actions ×4 | 23.2 | 24 |
| insight tabs | 58/67 × **22** | × 24 |
| name button | 82 × **18** | 24, no layout shift |

The first two are the interesting ones: they were never designed small, they were being squeezed below their own declared size. `flex-shrink: 0` restores what the CSS already asked for, and the sidebar toggles now render as proper squares rather than slivers.

Desktop layout confirmed unchanged by screenshot.

PR #1415, opened 2026-07-26
Improve guided intake and set-wise SQLite modeling

ai-christianson:
## Summary

- Keep homepage-created work in the active web conversation, even when email is the user's durable preferred contact.
- Restore selective consultant-style intake: broad assignments get bounded orientation and one useful native choice card; simple greetings still answer immediately.
- Make SQLite source modeling reliable in one shot. Structured results use an explicit current-batch boundary, unstructured interpretations use bound JSON rows, and older same-tool results remain queryable history without leaking into the active model context.
- Require exact known result links in the finished report while leaving unknown destinations plain.

## Root cause

The reproduced failures came from four independent gaps:

1. Homepage quick-create seeded the first turn from the durable preferred channel, so an active browser request could become an email conversation.
2. Removing mandatory Planning Mode also removed the selective intake contract. DeepSeek treated under-specified research jobs as ready to execute.
3. Same-tool source rows had no clean current-set boundary. Old events could leak URLs, IDs, and visible facts into the prompt, while imports ranged over unrelated historical calls.
4. Reconciliation detection mistook `result_id = excluded.result_id` in an UPSERT assignment for a source filter. The prompt therefore kept telling the model that SQLite reconciliation was still the next action after a successful import.

The fix adds a stored `source_batch_id` plus `is_current_batch`, scopes imports to the current batch/tool pair, removes stale same-tool source events from active prompt context while retaining them in SQLite, and limits reconciliation filter detection to actual predicates.

No planning state machine, output post-processor, behavioral retry loop, or channel heuristic was added.

## Verification

- 323/323 focused Django tests passed, including exact regressions for stale event leakage and the false post-UPSERT reconciliation loop.
- Complexity guardrails, test-tag validation, and `git diff --check` pass locally.
- Representative prompts are 566 bytes smaller than current `main` on first-run, explicit-send, and web-chat paths. Source LoC increases by 84 lines; production code is net +185 lines, with most of the PR devoted to tests/evals.

DeepSeek V4 Flash real-harness evidence:

- Guided-planning suite: 18/18 tasks — `92b59cb7-311c-40a7-bda8-84dbde7e8bbd`.
- Broad prospecting intake: 9/9 tasks across 3 runs — `3618a280-f13d-45d8-87b1-5547e127e4a1`.
- Simple greeting: 3/3 tasks across 3 runs — `3538e480-c92f-403f-9fea-ecfdc0e8818b`.
- Evidence-link owner report: 12/12 tasks across 3 runs — `0d246dbb-7f20-48f4-b7ba-b2ce322c518c`.
- Current-batch SQLite regression after the final fix: 9/9 tasks across 3 runs — `bd5a119b-0360-4c9d-bb75-53525d598730`.
- Full affected SQLite suite after the final fix: 38/48 (79.2%) — `03364531-3c53-4cba-bf6c-2b83402606cf`, up from 35/48 (72.9%) on the prior committed head. Remaining misses are existing DeepSeek over-query/auto-correction cases; the new current-batch scenario passed.

## Final gates

Exact head: `c9af45246a09d23aba0d2ffb2e7e05d83649d8b0`

- CI: running
- Preview deployment: pending
- Authenticated browser checks: pending

The PR will not be merged until exact-head CI and deployment pass and the preview proves:

- `Hi` responds quickly without research, SQLite, or an intake card.
- A broad homepage prospecting request stays on web and opens one useful native Insights card after bounded orientation.
- Continuing that task completes without SQLite errors/advisories and returns a structured report with evidence-backed links.


ai-christianson:
@codex review

PR #1417, opened 2026-07-27
Let a finished thought keep the room a streaming one gets

ai-christianson:
Ticket #405: planning steps in the timeline lack the thinking summary present in the detail view.

## What was wrong

While an agent is thinking, its reasoning renders over three lines (`.tool-cluster-live-preview__thinking-stream`, `max-height: calc(1.35em * 3)`). The instant the thought **completed**, it collapsed to the first clause of a single-line `0.6rem` caption.

So the thought visually degraded exactly when it became complete, and the planning step read as a label with nothing under it:

> **Planning next step**
> `PLAN-PROBE Let me process the current state. The user asked for a pipeline summary. I w…`

The full text was already being sent to the browser and already sat one click away in the detail popup. It simply wasn't shown.

Confirmed on **real staging data** before writing code: **137 of 380** timeline events carried non-empty reasoning, none of it visible in the timeline.

## The change

`deriveThinkingSummary` keeps the thought rather than its opening line — the previous helper took `firstMeaningfulLine`, so any multi-line reasoning lost everything after the first newline. It ends on a sentence boundary where one is available, so the summary reads as a thought rather than a truncation, and falls back to an ellipsis only when there's no sentence to end on.

It renders at the size and line count the streaming view already uses, so a thought no longer changes appearance the moment it finishes.

After:

> **Planning next step**
> `PLAN-PROBE Let me process the current state. The user asked for a pipeline summary. I will query the CRM table, filter to open opportunities, then summarise by owner before replying.`

One thing worth noting for review: the caption is rendered from **three** separate sites in `ToolClusterLivePreview`, and tagging only the first left the change invisible in the live DOM even though the summary text had already lengthened. All three are tagged now.

## Design

This is deliberately the quiet counterpart to the mood card from #404 — that one is a feeling shown at a glance, warm and hue-tinted; this one is a thought given enough space to actually be read, cool and text-forward. Together the timeline shows what an agent felt and what it was thinking.

## Verification

`clusterPreviewText.test.ts`, 7 tests, **confirmed red first** (helper absent → all fail). Covers the core regression — reasoning past the first line survives — plus sentence-boundary truncation, ellipsis fallback, markdown stripping, and non-thinking entries.

Rendered and inspected at 1440×900 and 390×844: summary wraps and ends on a complete sentence at both widths.

`tsc -b --force` clean · 51 test files / 285 tests green · CSS braces balanced 1555/1555.

### DIFF
--- a/frontend/src/components/agentChat/ToolClusterLivePreview.tsx
+++ b/frontend/src/components/agentChat/ToolClusterLivePreview.tsx
@@ -7,7 +7,7 @@ import { formatRelativeTimestamp } from '../../util/time'
 import { getFriendlyToolInfo, type FriendlyToolInfo } from '../tooling/toolMetadata'
 import { extractBrightDataSearchQuery } from '../tooling/brightdata'
 import { ToolIconSlot } from './ToolIconSlot'
-import { deriveSemanticPreview } from './tooling/clusterPreviewText'
+import { deriveSemanticPreview, deriveThinkingSummary } from './tooling/clusterPreviewText'
 import type { ToolClusterTransform, ToolEntryDisplay } from './tooling/types'
 import { parseToolSearchResult } from './tooling/searchUtils'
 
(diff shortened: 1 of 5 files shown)


ai-christianson:
Closing unmerged at Andrew's direction. The approach here is not the direction we want for #405, so the ticket goes back to unclaimed rather than sitting behind a stale PR. Nothing from this branch has landed on main.

PR #1420, opened 2026-07-27
Cancel stale MCP tool calls when newer human input arrives

matt-greathouse:
## Summary

- Cancel queued and in-flight tool calls when newer human input makes the agent run stale.
- Propagate stale-input checks through tool preparation and execution with clear cancellation reasons.
- Trigger remote MCP agent processing through the interactive enqueue path and preserve inbound generation tracking.
- Add regression coverage for stale tool cancellation and MCP-triggered processing.

## Testing

- Added and updated targeted unit tests for event processing and remote MCP behavior.
- Not run (not requested).


api/agent/core/event_processing.py
```diff
@@ -3859,12 +3907,18 @@ def _execute_prepared_tool_batch(
...
+        cancellation_reason = (
+            STALE_TOOL_CANCELLATION_REASON
+            if stale_prompt_checker and stale_prompt_checker()
+            else "Tool call was cancelled before execution because batch processing stopped."
+        )
         _cancel_unstarted_tool_calls(
             agent,
             prepared_batch.prepared_calls,
-            reason="Tool call was cancelled before execution because batch processing stopped.",
+            reason=cancellation_reason,
```
matt-greathouse, line 3957: `cancellation_reason` re-evaluates `stale_prompt_checker()` fresh inside `finally`, instead of capturing *why* the try block actually stopped. If the batch aborted for an unrelated reason (e.g. `_should_abort_processing` due to heartbeat loss / budget exhaustion) and newer human input happens to arrive in the small window before this `finally` runs, the cancelled tool calls get persisted with the misleading reason "newer human input arrived" even though that wasn't the actual cause of the abort. Conversely, a genuine mid-execution staleness break could theoretically report the generic "batch processing stopped" message if evaluated at a different instant.

This doesn't affect control flow (only the stored `result` message on the cancelled `PersistentAgentToolCall` rows), but it can mislead debugging/telemetry. Consider having `_execute_prepared_tool_batch_inner` return/propagate an explicit flag for "aborted specifically due to staleness" (e.g. reuse `execution_aborted`/a new field on `_ExecutedToolBatch`) rather than re-querying the checker after the fact.

---

tests/unit/test_event_processing.py
```diff
@@ -5358,6 +5359,146 @@ def _execute_side_effect(agent, prepared, **_kwargs):
         self.assertTrue(executed_batch.abort_after_execution)
         self.assertEqual(mock_execute_prepared_tool_call.call_count, 1)
 
+    def test_prepare_tool_batch_stops_when_newer_human_input_arrives(self):
```
matt-greathouse, line 5403: Good coverage for the "stale before any preparation" and "stale mid-execution" cases, but there's a gap: the new mid-*preparation*-loop break in `_prepare_tool_batch` (the `break` added around the per-call loop, distinct from the upfront early-return this test exercises) isn't directly covered — e.g. a batch with 2+ tool calls where `stale_prompt_checker()` flips to `True` only after the first call has already been prepared. Worth a test asserting `prepared_calls` contains the first call and `cancel_prepared_calls`/`abort_after_execution` are set, to lock in that partially-prepared calls still get cancelled correctly.

---

matt-greathouse (commented):
Reviewed the stale-tool-cancellation changes in `event_processing.py`, the `remote_mcp.py` trigger-processing fix, and the added test coverage.

**Overview**: This extends the existing `stale_prompt_checker` pattern (already used for streaming completions) to tool batch preparation/execution, so queued/in-flight tool calls get cancelled when newer human input bumps the inbound generation counter mid-run. It also fixes `_tool_send_agent_message` in the remote MCP service, which previously hardcoded `trigger_processing=False` and manually enqueued `process_agent_events_task.delay` directly — bypassing the generation-bump entirely, so MCP-triggered messages couldn't correctly signal staleness to a concurrently running agent loop. Routing through `enqueue_interactive_process_agent_events` also gets these messages onto the low-latency queue.

**Strengths**:
- New checks consistently follow the existing `_should_abort_processing` pattern (same `abort_after_execution`/`cancel_prepared_calls` fields, same break/return shape in both the parallel and serial execution paths).
- `_cancel_unstarted_tool_calls` correctly only flips `QUEUED` calls to `ERROR`, leaving already-completed calls untouched (verified by the "cancels remaining calls between calls" test).
- `_is_orchestrator_prompt_stale` closes over a fixed `accepted_human_generation` and compares against a monotonically increasing counter, so it can't flap during a single iteration.
- Good regression coverage added for the tool-batch cancellation paths and the remote MCP generation-bump behavior.

**Findings posted inline**:
1. `event_processing.py` — the cancellation `reason` string stored on cancelled tool calls is computed by re-invoking `stale_prompt_checker()` in the `finally` block rather than capturing the actual cause of the abort, which can misattribute the reason in edge cases. Doesn't affect control flow, just observability/debugging accuracy.
2. `tests/unit/test_event_processing.py` — missing direct test coverage for the mid-preparation-loop break (staleness detected after some, but not all, calls in a batch have been prepared).

No blocking issues found; both are minor/observability-level suggestions.

---

matt-greathouse (commented):
Re-reviewed after the new push (`e13fe8c` → `da4a4e8`). The new commit ("Preserve pending agent work across processing lock release") is a substantial addition on top of the original stale-tool-cancellation work — it replaces the plain pending-agent Redis `SET` with a coalescing model (generation + queue + generic-work flag) implemented via three atomic Lua scripts (`enqueue`/`claim`/`claim_many`), and proactively re-publishes pending work the moment the processing lock is released instead of waiting for the periodic drain sweep.

**Verified**:
- The enqueue script's queue-upgrade rule (never downgrade away from the interactive queue; always upgrade to it or on a newer generation) is internally consistent — traced through several call sequences without finding a case where it mis-prioritizes.
- The claim/claim-many scripts correctly clean up all four keys atomically and only pop entries actually present in the pending set.
- The in-memory fake Redis (`config/redis_client.py`) used by tests mirrors the real Lua scripts' logic faithfully — no divergence that would let a test pass against behavior the real script doesn't have.
- `enqueue_claimed_pending_work` correctly withholds `inbound_generation` for generic-work entries so they aren't skipped by the "already consumed" check, and the Celery-publish failure path in both `process_pending_agent_events_task` and the new lock-release path restores the claimed work (with a drain reschedule) rather than dropping it.
- Confirmed the project uses a single-node Redis (`redis.from_url`), so the multi-key `EVAL` calls aren't at risk of a cluster `CROSSSLOT` failure (this was worth double-checking given four distinct key names touched in one script).
- Test coverage for the new behavior is thorough: generation/queue coalescing, generic-work preservation, mutual exclusivity between the lock-release claim and the drain sweep, cleanup on `clear_processing_work_state`, and the broker-failure restore path all have dedicated tests.

No new blocking issues found in this commit. The two inline findings from the previous round (misleading cancellation-reason string in `_execute_prepared_tool_batch`'s `finally`, and a test-coverage gap for the mid-preparation-loop staleness break) are on code this commit didn't touch, so they still stand — as do the two `chatgpt-codex-connector` findings (non-retryable stale cancellations, and discarding stale completion state before returning).

---

PR #1427, opened 2026-07-27
Improve SQLite charter patch error guidance

matt-greathouse:
## Summary
- Provide actionable hints for incorrect `patch_text` argument counts.
- Explain SQLite incompatibility with PostgreSQL `E'...'` string literals.
- Preserve original SQL dialect feedback after autocorrection attempts.
- Add coverage for both error scenarios and confirm failed patches do not persist.

## Testing
- Added focused unit tests in `tests/unit/test_sqlite_batch.py`.


api/agent/tools/sqlite_batch.py
```diff
@@ -1610,6 +1621,16 @@ def _execute_with_autocorrections(
 def _get_error_hint(error_msg: str, sql: str = "") -> str:
     """Return a helpful hint for common SQLite errors."""
     error_lower = error_msg.lower()
+    if re.search(r"\bE\s*'", sql, re.IGNORECASE):
```
matt-greathouse: `\bE\s*'` fires on any string literal that simply ends in a space + the letter E — e.g. `'Vitamin E'`, `'Plan E'`, `'Grade E'`, `'Type E'` — not just Postgres `E'...'` escape syntax. Since this check runs first and unconditionally (no correlation with `error_msg`), a query containing one of these ordinary literals that fails for a *completely unrelated* reason (e.g. `no such column`, a typo, a real syntax error elsewhere) will get the misleading "SQLite does not support PostgreSQL E'...' literals" hint instead of the accurate one, which can send the agent chasing the wrong fix.

Repro:
```python
import re
re.search(r"\bE\s*'", "UPDATE t SET name='Vitamin E' WHERE id=1", re.IGNORECASE)  # matches
```

Consider also requiring the error message to actually look like a dialect/syntax problem (similar to how the backslash-quote check partially correlates with `error_lower`), e.g. only return this hint when `error_lower` contains something like `"syntax error"` or `"unrecognized token"` in addition to the regex match. The same regex is duplicated at the call site (sqlite_batch.py, the `hint_sql` selection a few lines below `_execute_sqlite_batch_inner`) and inherits the same false-positive risk.

---

matt-greathouse (commented):
Reviewed the charter-patch error guidance changes (`sqlite_batch.py`, `sqlite_guardrails.py`, tests).

**Overview**: makes `patch_text` variadic so it can report its own arg-count contract instead of a generic SQLite error, and adds a heuristic to detect Postgres `E'...'` escape-string literals and PostgreSQL-style `patch_text` misuse, preserving the original query text for hinting when autocorrection would otherwise obscure the dialect mistake. Tests cover both new error paths and confirm the failed patch doesn't persist.

**Issue found**: the new `\bE\s*'` regex used to detect Postgres escape-string literals is overly broad — it matches any ordinary string literal ending in "<space>E" (e.g. `'Vitamin E'`, `'Plan E'`, `'Grade E'`), not just the `E'...'` syntax. Because this check runs unconditionally before the other, more specific hints and isn't correlated with the actual error message, a query containing such a literal that fails for an unrelated reason will get a misleading hint. Left a line comment with a repro and a suggested fix (correlate with the error message, similar to the existing backslash-quote check).

Everything else (the variadic `patch_text` arg-count guard, the `patch_text`-specific hint gating on both SQL shape and error text, and the new tests) looks correct and follows existing conventions in the file.

---

PR #1429, opened 2026-07-27
Fix agent SQLite crashes and recover corrupted state safely

matt-greathouse:
## Summary
- Prevent agent crashes caused by SQLite state and batch-processing failures
- Add SQLite recovery and file-handling safeguards across agent tools and sandbox compute
- Expand unit coverage for event processing and SQLite state recovery

## Testing
- Added and updated focused unit tests for event processing and SQLite recovery behavior
- Not run (not requested)


api/agent/tools/sqlite_state.py
```diff
@@ -1012,24 +1026,20 @@ def _restore_sqlite_db_from_storage(storage_key: str, db_path: str, agent_uuid:
...
             agent_uuid,
             exc_info=True,
         )
-        try:
-            if os.path.exists(db_path):
-                os.remove(db_path)
-        except Exception:
-            logger.debug("Failed to delete partially restored SQLite DB for agent %s", agent_uuid, exc_info=True)
-        return False
+        raise
```
matt-greathouse, line 1073: Restore-time corruption now has no self-healing path. Previously, a failed restore fell back to a fresh DB (silent data loss, but the agent stayed functional). Now it re-raises, and at this point in `_agent_sqlite_db_uncoordinated` no `SQLiteStateSession`/checkpoint has been created yet (that only happens a few lines below), so there is nothing to recover *from*.

Since the persisted blob in storage is never repaired or discarded here, every future call to `agent_sqlite_db(...)` for this agent will hit this same `raise` again — the outer tool-exception handler in `event_processing._execute_prepared_tool_call` will catch it and return a generic "Tool execution failed" message, but the agent's SQLite capability is now permanently disabled with no automatic recovery, unlike the mid-run corruption cases this PR does handle (which self-heal via the checkpoint). Worth deciding whether restore-time corruption should also attempt one automatic "start fresh" fallback (perhaps behind a distinct error code/alert) rather than failing forever.

---

api/agent/tools/sqlite_state.py
```diff
@@ -1046,102 +1056,218 @@ def agent_sqlite_db(agent_uuid: str):  # noqa: D401 – simple generator context
...
+                        recovery_exc,
+                        recovered=False,
                     )
+                    raise
+                _log_sqlite_persistence_error(
+                    agent_uuid,
+                    exc,
+                    recovered=session.recovery_count > recovery_count_before,
+                )
+                raise
```
matt-greathouse, line 1152: This `raise` fires for *any* `SQLiteStateError` from `_persist_validated_sqlite_state`, including a purely transient upload failure (network blip, storage throttling) where the local DB is perfectly valid — `_upload_sqlite_archive` wraps generic `Exception` from the storage backend into `SQLiteStateValidationError` too. Previously this path only logged (`logger.exception(...)`) and let the agent step continue.

Now a transient storage error aborts whatever was using `agent_sqlite_db(...)` (an LLM loop iteration or a `sqlite_batch` call), even though `session.protect(phase="final_persistence_failure")` just re-validated that the local state is fine. Given cloud storage hiccups aren't rare, this could noticeably raise the agent's error rate for non-corruption reasons. Consider only hard-failing on genuine corruption and treating a validated-but-unuploaded state as retryable/best-effort instead.

---

api/agent/core/event_processing.py
```diff
@@ -6996,6 +7011,14 @@ def _token_usage_fields(token_usage: Optional[dict], response: Any) -> dict:
                     logger.debug("Failed to resolve web stream target for agent %s", agent.id, exc_info=True)
 
                 try:
+                    recovered = protect_current_sqlite_state(
+                        phase="before_llm_completion",
+                        checkpoint=True,
+                    )
```
matt-greathouse, line 7096: This does a full checkpoint (`create_validated_sqlite_snapshot` → backup + validate + atomic replace) unconditionally on **every** agent loop iteration before every LLM call, even iterations where no SQLite write happened. `sqlite_batch` already checkpoints after its own execution (`_sqlite_state_guard_result(phase="sqlite_batch_after_execution", checkpoint=True)` in `sqlite_batch.py`), so on iterations following a batch call this re-snapshots state that hasn't changed since. For agents with a large SQLite state file this adds a full-DB-sized I/O op to the hot path of every single completion, not just ones that touched SQLite. Might be worth tracking a "dirty since last checkpoint" flag so this can skip the redundant snapshot.

---

api/agent/tools/sqlite_recovery.py
```diff
@@ -0,0 +1,176 @@
...
+        initialize_sqlite_file(self.db_path)
+
+    @contextlib.contextmanager
+    def _observe(self, *, operation: str, phase: str):
+        started_at = time.monotonic()
+        error = None
+        with tracer.start_as_current_span(f"{operation.title()} Agent SQLite State") as span:
+            try:
+                yield
+            except (OSError, sqlite3.Error, SQLiteStateError) as exc:
```
matt-greathouse: Minor: `error` is only set for `(OSError, sqlite3.Error, SQLiteStateError)`. If the wrapped operation ever raises something outside that set (e.g. a stray `TypeError`/`AttributeError` from a future bug), it still propagates correctly, but the `finally` block below will log/trace it as `sqlite.validation.ok: True` / `outcome: "succeeded"` since `error` stays `None` — a misleading signal in traces/logs for exactly the kind of unexpected failure you'd most want visibility into.

---

matt-greathouse (commented):
Reviewed the SQLite crash/recovery hardening. The new checkpoint-and-recover mechanism (`sqlite_recovery.py`, `sqlite_files.py`) and its test coverage are solid — the mid-run corruption recovery path is well tested (single-recovery-attempt safety valve, checkpoint-IO-failure-doesn't-roll-back-valid-state, etc.).

Left 4 comments, mainly around the edges of the new "fail loud instead of silently starting fresh" philosophy:
- **Restore-time corruption has no self-healing path** and will permanently disable SQLite for that agent (no checkpoint exists yet to recover from) — the one corruption scenario this PR doesn't actually fix.
- **Persistence-phase failures now hard-fail even for transient storage/network errors**, not just genuine corruption, which could raise the agent's error rate for non-corruption reasons.
- **Checkpointing before every LLM completion** duplicates the checkpoint `sqlite_batch` already does after execution, adding redundant full-DB I/O to the hot path.
- Minor observability nit in `_observe()`'s exception classification.

None of these are blocking bugs — the recovery mechanism itself works as designed and is well tested — but worth a look before merge.

---

PR #1446, opened 2026-07-28
Bound stalled model streams so agent queues recover

ai-christianson:
## Summary

- Enforce the existing 30-second stream idle bound between every LLM chunk, not only before the first chunk.
- Fail over immediately after a stalled stream instead of replaying the same provider non-streaming.
- Preserve existing lease finalization: if providers fail, release processing state and enqueue exactly one coalesced pending turn.

## Production evidence

Three independent production agents reproduced one mechanism behind internal bugs #488 and #455. Each persisted a tool result, rebuilt its prompt, began a streamed completion, then held the per-agent processing lock while schedules and inbound messages queued. The Keith incident had no next provider chunk for 8m44s; the stale-generation check ran and the lock released only when another chunk finally arrived. Megan and Sophie showed the same first-chunk-then-silence signature on separate routes. `planning_state=skipped` was correlation, not the cause.

## Tests

- Red before fix: the second-chunk timeout regression failed because no timeout was raised.
- `uv run python manage.py test --tag=batch_event_llm --settings=config.test_settings` — 99 passed.
- `uv run python manage.py test tests.unit.test_event_processing_locking tests.unit.test_event_processing_pending_drain tests.unit.test_event_processing_pending_followup --settings=config.test_settings` — 18 passed.
- Complexity budget passes. Counted source grows by 23 lines, 262034 to 262057; explicit baseline bump included.

No LLM eval was added: this is a deterministic transport/lifecycle defect, and a stochastic real-model eval cannot inject or prove an inter-chunk network stall.

## Scope

Root-cause fix at the LLM streaming transport boundary, plus verification of existing queue finalization. Internal #409 (outbound metadata) and #490 (peer-DM tool choice) are unrelated and intentionally excluded.

### DIFF
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -59,7 +59,13 @@ from .processing_flags import (
     processing_lock_storage_keys,
     set_processing_heartbeat,
 )
-from .llm_utils import EmptyLiteLLMResponseError, raise_if_empty_litellm_response, raise_if_invalid_litellm_response, run_completion
+from .llm_utils import (
+    EmptyLiteLLMResponseError,
+    StreamIdleTimeout,
+    raise_if_empty_litellm_response,
+    raise_if_invalid_litellm_response,
+    run_completion,
+)
 from .multimodal_context import collect_fresh_read_file_image_attachments, prepare_multimodal_read_file_request
 from .llm_streaming import StreamAccumulator
 from .tool_arg_streaming import ChatBodyStreamExtractor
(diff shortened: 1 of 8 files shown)


ai-christianson:
Final verification on exact head `6131cc7cc10aeb856db3c36ea3094aa01f8b95f8`:

- Full CI green: all 10 backend shards, frontend, sandbox, tag, combined result, and complexity guardrails.
- Preview deploy workflow succeeded; web, interactive worker, worker, beat, and Discord pods all ran the exact head image.
- Authenticated API ping and MCP initialize/tool discovery succeeded on `pr-1446.ship.gobii.ai`.
- Contained DeepSeek V4 Flash probe completed a real streamed turn; debug diagnostics showed `processing_active: false` and zero recent errors afterward.
- Authenticated headed Chrome rendered the agent chat page successfully; ticket fragment cleared and no login form remained. Screenshot visually inspected.
- Probe agent archived after validation.

This deterministic lifecycle defect is covered by red-to-green unit/integration regressions; an LLM judge eval would not inject or prove a transport stall.

PR #1451, opened 2026-07-28
Add structured peer messages and handoff boundaries

matt-greathouse:
## Summary

- Add structured payloads for peer-agent messages and handoffs
- Preserve and render peer metadata across compaction, timelines, and agent chat
- Prevent duplicate outbound peer messages
- Add responsibility-boundary and structured-handoff eval scenarios

## Testing

- Added and updated targeted unit tests for peer messaging, payloads, timeline rendering, event processing, and SQLite message storage
- Added eval coverage for structured peer handoffs and responsibility boundaries


api/agent/core/compaction.py
```diff
@@ -238,9 +242,12 @@ def llm_summarise_comms(
...
         role = "Assistant" if msg.is_outbound else "User"
-        # Truncate each message body to avoid huge token usage (4k chars cap).
-        body_preview = msg.body[:4000]
-        lines.append(f"{role}: {body_preview}")
+        content = msg.body or ""
+        payload = get_structured_peer_payload(msg.raw_payload)
+        if payload is not None:
+            content = f"{content}\nStructured payload:\n{canonicalize_structured_peer_payload(payload)}"
+        # Keep the body and structured data under the same per-message compaction cap.
+        lines.append(f"{role}: {content.strip()[:4000]}")
```
matt-greathouse: The prose body and the structured payload now share a single 4000-char truncation window. A structured payload can be up to 64 KB (`STRUCTURED_PEER_PAYLOAD_MAX_BYTES`), so for any non-trivial payload this will chop the JSON mid-string, handing the summarization LLM invalid/garbled JSON with no indication that it was truncated. Consider truncating body and payload independently (e.g. cap each separately, or truncate the payload at a JSON-safe boundary) rather than truncating the combined string.

Also, `llm_summarise_comms` doesn't appear to have test coverage for the structured-payload path (only `test_peer_structured_payload.py`/`test_event_processing.py` cover the prompt-building side, not compaction) — worth adding a case with a large payload to lock in the desired truncation behavior.

---

api/agent/tools/outbound_duplicate_guard.py
```diff
@@ -303,6 +310,7 @@ def detect_recent_duplicate_message(
         conversation_id: Optional[UUID] = None,
         similarity_threshold: Optional[float] = None,
         exact_only: bool = False,
+        structured_payload: StructuredPeerPayload | None | object = _NO_STRUCTURED_PAYLOAD,
```
matt-greathouse: Minor: the type hint `StructuredPeerPayload | None | object` is redundant — `object` already subsumes `StructuredPeerPayload | None`, so this reads as documentation but is a bit misleading about the sentinel's actual role (distinguishing "not passed" from "passed as None"). Might be clearer as `Any` with a short comment noting `_NO_STRUCTURED_PAYLOAD` is a distinct sentinel from `None`.

---

api/agent/core/prompt_context.py
```diff
@@ -4448,15 +4475,20 @@ def _build_sqlite_messages_snapshot_records(
             continue
 
         body = _redact_signed_filespace_urls(message.body or "", agent)
-        body_bytes = len(body.encode("utf-8"))
-        if total_body_bytes + body_bytes > max_total_body_bytes:
+        raw_payload = message.raw_payload if isinstance(message.raw_payload, dict) else {}
+        structured_payload = get_structured_peer_payload(raw_payload)
```
matt-greathouse, line 4476: `get_structured_peer_payload`/`canonicalize_structured_peer_payload` are computed here for the byte-budget check and then recomputed from scratch inside `_build_message_sqlite_record` (~line 4446) for the same message/raw_payload. Not a correctness issue, just duplicated work — could thread the already-computed `payload_json` through `selected_messages` instead of recalculating it downstream.

---

matt-greathouse (commented):
Reviewed the structured peer-message payload feature (`api/agent/structured_peer_payload.py` plus its integration into peer messaging, compaction, prompt context, SQLite snapshotting, timeline serialization, and the React chat card).

Overall this is well-scoped and thoroughly tested (unit tests for validation, duplicate detection, SQLite storage, timeline serialization, and a new real-harness eval suite). The `_source == PEER_DM_SOURCE` guard in `get_structured_peer_payload` correctly keeps this feature scoped to peer DMs only, and the duplicate-detection changes correctly preserve existing body-only behavior while adding payload-aware exact matching.

Three non-blocking findings, most notable first:

1. **`compaction.py`**: body and structured payload share one 4000-char truncation cap, so a large (up to 64 KB) payload can be truncated mid-JSON with no indication, feeding invalid JSON into the summarization prompt. No test covers this path.
2. **`outbound_duplicate_guard.py`**: a redundant/misleading type hint (`X | None | object`) on the new sentinel parameter — style only.
3. **`prompt_context.py`**: structured payload is computed twice for the same message across two functions — a minor duplication, not a bug.

Nothing here blocks merging; the core logic (validation, size limits, duplicate detection, empty-body handling, frontend rendering) all checked out correctly against the test suite.

---

PR #1455, opened 2026-07-28
Route managed onboarding check-ins to peer managers

matt-greathouse:
## Summary

- Clarify prompt instructions so managed agents preserve charter reporting boundaries and route routine check-ins to named peer managers.
- Prevent unsolicited owner welcomes or manager introductions during first-run processing.
- Add a real-harness eval covering onboarding, manager routing, schedule control-plane state, and charter preservation.
- Expand responsibility-boundary unit coverage for the new prompt behavior and scenario registration.

## Testing

- Added focused unit tests for prompt precedence and managed-agent first-run behavior.
- Added a core real-harness eval for managed onboarding check-ins.


api/evals/tests/test_responsibility_boundaries.py
```diff
@@ -54,7 +60,7 @@ def test_peer_contract_is_compact_and_ownership_first(self):
         self.assertIn("Never relay shared-channel requests by DM", instruction)
         self.assertIn("Synthesize only owned, attributed work", instruction)
         self.assertNotIn("freely", instruction)
-        self.assertLessEqual(len(instruction.split()), 95)
+        self.assertLessEqual(len(instruction.split()), 205)
```
matt-greathouse: This test previously guarded prompt conciseness for `_get_peer_communication_instruction` (95-word cap). The new paragraph brings the instruction to 201 words, and the cap was raised to 205 to match — a 4-word margin. At that point the assertion isn't really enforcing compactness anymore, just confirming the new text was pasted in. Worth either trimming the new paragraph (see the duplication note in `prompt_context.py`) or picking a more principled ceiling than "current length + a few words," since this instruction is injected into every prompt for an agent with peer links and adds directly to per-call token cost.

---

api/agent/core/prompt_context.py
```diff
@@ -3909,7 +3909,24 @@ def _get_peer_communication_instruction() -> str:
...
+        "plane handled state. Use the account owner only if the charter requires it, the manager escalates, or a material "
+        "team decision is blocked.\n"
+    )
+
+
+def _get_managed_peer_first_run_instruction() -> str:
+    return (
+        "FIRST-RUN RECIPIENT PRECEDENCE: The preferred owner contact is fallback, not mandatory. If the Current "
+        "Charter routes routine coordination to a named reachable peer manager, Route 1 above does not authorize an "
+        "owner welcome. Do not send an untriggered manager introduction either; sleep until assigned work or a "
```
matt-greathouse: `_get_managed_peer_first_run_instruction` (lines 3924-3930) largely restates the new paragraph added just above in `_get_peer_communication_instruction` (lines 3913-3921) — both say, in slightly different words, "if the charter routes to a named peer manager, don't send an owner welcome or an unsolicited manager introduction." Since both blocks fire together for a managed agent's first run (`_get_peer_communication_instruction` is gated on `has_peer_links` and included in the critical group, and this first-run addendum is gated on the same peer-link check a few lines below), the prompt ends up repeating the same directive twice. Consider having the first-run addendum only state what's specific to the first-run case (that the welcome-message instruction above doesn't override this) and drop the restated peer-manager/no-introduction rule, to save tokens on every first-run cycle for managed agents.

---

api/agent/core/prompt_context.py
```diff
@@ -4214,6 +4231,10 @@ def _get_system_instruction(
                 "\n\n"
                 + _get_first_run_welcome_message_instruction(welcome_target=welcome_target)
             )
+            if AgentPeerLink.objects.filter(is_enabled=True).filter(
+                Q(agent_a=agent) | Q(agent_b=agent)
+            ).exists():
```
matt-greathouse: This `AgentPeerLink.objects.filter(is_enabled=True).filter(Q(agent_a=agent) | Q(agent_b=agent)).exists()` query duplicates the identical `has_peer_links` check already computed earlier in the prompt-building pipeline (around line 1911, in the critical-group builder). Since that result isn't threaded through to `_get_system_instruction`, this adds a third redundant DB round-trip per prompt build for agents with peer links. Low-severity given it's bounded to the first-run path, but if `has_peer_links` (or the agent's peer links) is already available/computable once upstream, passing it in would avoid the extra query.

---

matt-greathouse (commented):
Reviewed the diff and cross-checked the changed code against the production paths it touches (prompt building in `prompt_context.py`, and the schedule/cron-trigger helpers the new eval scenario exercises in `agent_schedules.py` / `process_events.py`).

**Correctness:** solid. The new `ManagedOnboardingCheckinScenario` faithfully mirrors production behavior — the `PersistentAgentCronTrigger` construction matches `process_agent_schedule_trigger_task` exactly, the `claim_schedule_occurrence` revision/enabled/next_run_at expectations match its actual implementation, and the `PersistentAgentMessage`/`AgentPeerLink` queries use real, indexed model fields. Ran the full `api.evals.tests.test_responsibility_boundaries` suite locally — all 20 tests pass, including the new ones.

**Findings posted inline** (all low severity, no blockers):
1. The word-count guard on `_get_peer_communication_instruction` was loosened from 95→205 words to fit the new paragraph (which is itself 201 words) — it no longer meaningfully guards conciseness.
2. The new first-run addendum (`_get_managed_peer_first_run_instruction`) duplicates most of the directive already stated in the paragraph added to `_get_peer_communication_instruction`, and both fire together on a managed agent's first run.
3. A third, redundant `AgentPeerLink` existence query was added to `_get_system_instruction`, duplicating a check already done earlier in the prompt pipeline.

None of these block merging — they're prompt-token/maintainability nits rather than functional bugs.

---

PR #1457, opened 2026-07-28
Add MCP message sending with outbound routing and duplicate protection

matt-greathouse:
## Summary

- Add support for sending messages through MCP servers.
- Route outbound MCP messages with source metadata and duplicate-send protection.
- Improve tool argument streaming, runtime handling, and daily-limit behavior.
- Update agent chat timeline rendering, message event types, documentation, and eval coverage.

## Testing

- Added and updated unit tests for MCP replies and tool argument streaming.
- Updated message quality eval scenarios and CI registration.
- Added frontend coverage for message event rendering.


api/agent/tools/mcp_sender.py
```diff
@@ -0,0 +1,167 @@
...
+                        ),
+                    },
+                },
+                "required": ["body", "will_continue_work"],
+            },
+        },
+    }
+
+
+@handle_link_reference_errors
```
matt-greathouse, line 84: `send_mcp_message` is missing from `_EMBEDDED_FIELDS` in `api/agent/core/link_references.py:28`, so any MCP reply body containing a `$[link:...]` reference is rejected instead of passed through.

Trace through `resolve_link_reference_params` (called at `event_processing.py:3240` before dispatch) with `tool_name="send_mcp_message"`:

- `_allowed = _embedded_fields("send_mcp_message", params)` → `set()`, because the tool isn't a key in `_EMBEDDED_FIELDS`
- for the `body` value, `_path[0] in _allowed` → `False` (falls through instead of returning the value untouched)
- `_REFERENCE_RE.fullmatch(stripped)` → `None` for any body that is more than a bare token
- `tool_name and _contains_reference_syntax(value)` → `True` → raises `LinkReferenceResolutionError`

Caught at `event_processing.py:3241`, so it isn't a crash — but the agent gets back `"Query not executed: link references are unsupported in send_mcp_message.body."`

Concretely: body `Here is the report: [Q3 summary]($[link:LABCD1234EFGH5678])` errors out on `send_mcp_message`, while the byte-identical body succeeds on `send_chat_message` (which has `"send_chat_message": "body"` registered). Every other message tool — `send_email`, `send_sms`, `send_agent_message`, `send_discord_message` — is registered too; this is the only send tool that isn't.

There's a second, subtler divergence: a body consisting of *only* a bare `$[link:L...]` token doesn't error, but takes the `exact and tool_name not in _STRICT_TOOLS` branch and gets **expanded into a raw URL**, since `_STRICT_TOOLS` is derived from `_EMBEDDED_FIELDS` and therefore also excludes `send_mcp_message`. So the tool silently behaves two different ways depending on whether the reference is alone in the body.

Applying `@handle_link_reference_errors` here signals the intent was for references to flow through, but the decorator only converts the raise into an error response — it doesn't make references work. The fix is the registration:

```python
_EMBEDDED_FIELDS = {..., "send_chat_message": "body", "send_mcp_message": "body", ...}
```

which fixes both behaviors at once (`_STRICT_TOOLS` picks it up automatically). Worth a test asserting an MCP body with an embedded link reference round-trips unmodified — the current `tests/unit/test_mcp_reply.py` has no link-reference case, which is why this passes CI.

Given the PR's framing around delivering decision-ready MCP results, replies that can't carry links seem like a meaningful gap.

---

api/agent/core/daily_limit_mode.py
```diff
@@ -9,6 +9,7 @@
         "send_email",
         "send_sms",
         "send_chat_message",
+        "send_mcp_message",
```
matt-greathouse, line 12: `send_mcp_message` was added to the `CREDIT_MESSAGE_TOOL_NAMES` set, but `CREDIT_MESSAGE_ONLY_ALLOWED_TOOL_NAMES_TEXT` just below (line 20) still enumerates the old list:

```python
CREDIT_MESSAGE_ONLY_ALLOWED_TOOL_NAMES_TEXT = (
    "send_email, send_sms, send_chat_message, send_discord_message, send_agent_message, "
    "and sleep_until_next_trigger"
)
```

That string is the human-readable contract shown to the model in two places:

- `prompt_context.py:3113` — the `credit_message_only_mode` prompt section: *"Only message and sleep tools are available right now: {TEXT}. Do not attempt any other tools or non-message work."*
- `event_processing.py:3563` — the correction step persisted when a disallowed tool is called

So in credit-message-only mode the tool is *functionally* permitted (`filter_tools_for_credit_message_only_mode` keeps it, and `is_credit_message_only_allowed_tool("send_mcp_message")` returns `True` — the new test at `test_mcp_reply.py:1251` asserts exactly this), but the agent is *told* it isn't in the allowed set.

The failure mode: an agent servicing an MCP request while out of credits reads an allow-list that omits `send_mcp_message` and contains `send_chat_message`, and is steered toward a human channel to reply. That's the precise behavior `RemoteMcpNoHumanContactScenario` in this PR was added to guard against — but that scenario doesn't exercise the credit-exhausted path, so it won't catch the regression.

Add `send_mcp_message` to the text. Better still, derive it from the set so the two can't drift again:

```python
CREDIT_MESSAGE_ONLY_ALLOWED_TOOL_NAMES_TEXT = (
    ", ".join(sorted(CREDIT_MESSAGE_TOOL_NAMES)) + ", and sleep_until_next_trigger"
)
```

---

api/agent/core/event_processing.py
```diff
@@ -6854,7 +6869,11 @@ def _refresh_inbound_routing_scope(generation: int) -> bool:
                 routing_profile = get_current_eval_routing_profile()
                 prefer_low_latency = iteration_prefers_low_latency
                 feedback_inbound = get_current_inbound_message(agent)
-                iteration_tools = _filter_incompatible_reply_tools(iteration_tools, feedback_inbound)
+                iteration_tools = _filter_incompatible_reply_tools(
+                    iteration_tools,
+                    feedback_inbound,
+                    mcp_available=get_recent_mcp_inbound_message(agent) is not None,
```
matt-greathouse: This runs a DB query on every pass of the agent tool loop, and it fetches more than it needs.

`get_recent_mcp_inbound_message` (`routing.py:120`) does `.select_related("conversation", "from_endpoint")` and `.first()` — materializing a full message row plus two FK joins — when the call site only needs the boolean `is not None`. It also filters on `raw_payload__source_kind="mcp"`, a JSON key lookup that is unlikely to be index-backed, over the agent's whole message history within the retention window.

Two cheap improvements:

```python
mcp_available=agent_has_recent_mcp_inbound(agent)   # .exists() variant, no select_related
```

and hoisting it out of the loop. The result is near-constant for the duration of a run — it can only flip if a new inbound MCP message lands mid-loop, and the surrounding code already has an explicit mechanism for that case (`_current_human_inbound_generation()` / `_refresh_inbound_routing_scope`) rather than re-querying unconditionally each iteration.

The full-object fetch is still needed inside `execute_send_mcp_message` (`mcp_sender.py:82`), which uses `inbound.conversation`, `inbound.from_endpoint`, and `inbound.id` — so an `.exists()` helper alongside the existing getter, rather than a change to it, is the smaller edit.

Minor — no correctness impact, just avoidable per-iteration work in a hot path.

---

matt-greathouse (commented):
## Review: MCP message sending with outbound routing and duplicate protection

Reviewed the full diff (23 files, +848/-44). The design is coherent: `send_mcp_message` persists a `DELIVERED` reply parented to the newest MCP-origin inbound message, reusing the existing web conversation as a transport while systematically suppressing every human-facing side effect of an outbound message — owner notifications, unread state, signup-preview transitions, and implied-send context. The isolation looks thorough; I traced each suppression point and they line up.

Verified as correct along the way:
- `WEB_SESSION_RETENTION_DAYS` and `WEB_CHAT_MESSAGE_MAX_LENGTH` both exist in `config/settings.py`
- The duplicate guard's `source_kind` filter composes correctly with the existing conversation/address filters; `from_endpoint__channel` matches the web endpoint the MCP reply is written through
- The production MCP path (`remote_mcp.py:1206`) uses `CommsChannel.WEB` + `source_kind="mcp"`, matching what the eval/test scaffolding injects — so the tests exercise the real shape
- `raw_payload` is `default=dict` and non-nullable, so the `isnull` leg of the new `latest_visible_outbound_message_queryset` filter is redundant but harmless
- Credit exemption is consistent — `_ensure_credit_for_tool` short-circuits, making the `TOOL_CREDIT_NO_CHARGE_ON_ERROR_TOOLS` omission a non-issue

Test coverage is good: 290 lines of unit tests covering persistence/correlation, staleness, newest-conversation selection, duplicates, attachments, and notification suppression, plus frontend direction tests and a new real-harness eval.

### Findings

Three comments inline, one of which I'd want addressed before merge:

1. **`send_mcp_message` missing from `_EMBEDDED_FIELDS`** (`mcp_sender.py`) — MCP reply bodies containing `$[link:...]` references are rejected with an error, while identical bodies succeed on `send_chat_message`. A bare-token body behaves differently again (silently expanded to a raw URL). This is the only send tool not registered there. No test covers it, which is why CI is green.

2. **`CREDIT_MESSAGE_ONLY_ALLOWED_TOOL_NAMES_TEXT` not updated** (`daily_limit_mode.py`) — the tool is functionally allowed in credit-message-only mode but the prompt text tells the model it isn't, steering it toward `send_chat_message` for an MCP reply. That's the exact behavior the new `RemoteMcpNoHumanContactScenario` guards against, in a path that scenario doesn't cover.

3. **Per-iteration query** (`event_processing.py`) — `get_recent_mcp_inbound_message` runs each loop pass with two wasted `select_related` joins when only existence is checked. Minor.

Note that `mergeable_state` is currently `unstable` — worth confirming CI is green on the new `mcp_reply_batch` tag added to `shard_04`.

---

scripts/budgets/complexity_budgets.json
```diff
@@ -149,6 +149,6 @@
       ".yml",
       ".zsh"
     ],
-    "limit": 262609
+    "limit": 280000
```
matt-greathouse, line 152: This hand-edit works around the budget check rather than re-baselining it, and history suggests it won't survive.

`check_complexity_budgets.py:474` sets the limit to the *exact* measured count:

```python
budget["source_loc"]["limit"] = loc.line_count
```

So the budget is a strict ratchet with zero headroom by design. `280000` is ~16.6k lines above the post-PR actual (~263.4k) — roughly 6% of silent headroom that would absorb future growth without any PR having to acknowledge it.

Two concrete problems:

**1. It gets reverted.** `280000` has been hand-set three times already, and each time the next `--update-baselines` run stomped it back to the tight value:

| hand-set to 280000 | reset by | reset to |
|---|---|---|
| `0030247` | `c725137` | 262223 |
| `c21ac34` | `4791638` | 262223 |
| `569c073` | `c1783d4` | 262444 |

Main's own history tracks tightly — 261920 → 261939 → 262034 → 262057 → 262088 → 262223 → 262444 → 262609. This edit is the fourth instance of the same churn.

**2. `baseline_sha` is now inconsistent with `limit`.** The commit changed only `limit`, leaving `baseline_sha` at `9785d06…` (carried in from main's last regeneration in `aad79f2`). But `_budget_metadata()` at line 437 writes `baseline_sha` and `generated_by` together with the limit, so the file still self-describes as:

```json
"generated_by": "uv run python scripts/check_complexity_budgets.py --update-baselines"
```

…while carrying a limit that command would never produce. The metadata now misrepresents how the value was derived.

Suggested fix — regenerate instead of hand-editing, so `limit` and `baseline_sha` move together:

```
uv run python scripts/check_complexity_budgets.py --update-baselines --loc-only
```

(the script takes a `loc_only` flag at line 468). That lands the limit at the true post-PR count and keeps the ratchet meaningful for the next PR.

---

matt-greathouse (commented):
## Follow-up: re-reviewed after force-push (`0d7e36d` → `dad138c`)

The branch was force-pushed shortly after my [first review](https://github.com/gobii-ai/gobii-platform/pull/1457#pullrequestreview-4800024981). I re-checked the delta so the earlier findings don't get lost in the rebase.

**The rebase is clean.** The original commit was rebased onto `b1b1204` (the merge of #1456). Since #1456 also touches `prompt_context.py`, there was real conflict potential there — I diffed this PR's own delta to that file before and after the rebase and it's byte-identical, same hunks at the same line numbers. No conflict damage.

**All three earlier findings still stand**, unmodified by the rebase — `mcp_sender.py`, `daily_limit_mode.py`, and `event_processing.py` are byte-identical across the two heads, so those inline comments remain correctly anchored:

1. `send_mcp_message` missing from `_EMBEDDED_FIELDS` — MCP replies containing `$[link:...]` are rejected (the one I'd fix before merge)
2. `CREDIT_MESSAGE_ONLY_ALLOWED_TOOL_NAMES_TEXT` not updated — steers MCP replies onto a human channel when credits are exhausted
3. Per-iteration `get_recent_mcp_inbound_message` query with two wasted joins — minor

**The only new content** is the one-line `source_loc` budget bump in `dad138c`. Comment inline — short version: it's a hand-edit where the tooling expects a regeneration, it leaves `baseline_sha` inconsistent with `limit`, and the same `280000` override has been reverted by `--update-baselines` three times before. Non-blocking, but `--update-baselines --loc-only` would be the durable form.

CI is still `unstable` on this head.

---

PR #1466, opened 2026-07-29
Add bot detection check page and identity signal verification

matt-greathouse:
## Summary

- Add a bot detection check page and supporting styles and JavaScript.
- Integrate identity and fingerprint signals into account authentication flows.
- Add configuration, routing, templates, context helpers, and service support.
- Cover bot-check behavior and page integrations with unit tests.

## Testing

- Added and updated unit tests in `tests/unit/test_bot_check.py` and `tests/unit/test_pages.py`.


pages/bot_check.py
```diff
@@ -0,0 +1,1253 @@
...
+            "hardware",
+            "Hardware profile",
+            "info",
+            (
+                f"{int(client['hardware_concurrency'])} logical processors"
+                if client.get("hardware_concurrency") is not None
+                else "Processor count unavailable"
+            ),
+            (
+                f"Reported memory: {client.get('device_memory')} GB; "
```
matt-greathouse: `client.get('device_memory')` has no fallback for missing values, unlike the adjacent `color_depth` check on the same block which falls back to `'unknown'`.

`navigator.deviceMemory` is unsupported in Safari and Firefox, so `bot_check.js` sends `device_memory: null` for most non-Chromium visitors. That renders literally as "Reported memory: None GB" in both the shown report and the copyable JSON.

Suggest: `client.get('device_memory') or 'unknown'` (or similar) to match the pattern used elsewhere in this function.

---

pages/bot_check.py
```diff
@@ -0,0 +1,1253 @@
...
+    if fingerprint_client["integration_error"]:
+        fingerprint_status = (
+            "client_error"
+            if fingerprint_client["integration_error"] == "agent_error"
+            else f"client_{fingerprint_client['integration_error']}"
+        )
+    elif fingerprint_config["enabled"] and not event_id:
+        fingerprint_status = "missing_event"
+
+    if event_id and fingerprint_config["enabled"]:
```
matt-greathouse, line 1210: This `if event_id and fingerprint_config["enabled"]:` block unconditionally overwrites `fingerprint_status`, even when the preceding block (line 1206) already set a `client_*` status reflecting a reported client-side integration error.

`bot_check_complete` is a public POST endpoint that trusts `payload.fingerprint_client` and `payload.fingerprint_event_id` as given. If a request includes both a non-empty `fingerprint_event_id` and a `fingerprint_client.integration_error` (e.g. `csp_block`), the status set at line 1207-1211 (`client_csp_block`) gets clobbered here to `browser_only`/`unavailable`, losing the diagnostic signal that a client-side failure was actually detected.

Consider guarding this block with `elif` or checking whether `fingerprint_client["integration_error"]` was already set before overwriting.

---

static/js/account_identity_signals.js
```diff
@@ -101,24 +101,69 @@
...
+    }
+    return "agent_error";
+  }
+
   function createFpjsPromise(options) {
     return withTimeout(
       import(options.loaderUrl)
-        .then(({ load, defaultEndpoint }) => {
+        .then((Fingerprint) => {
+          if (typeof Fingerprint.start === "function") {
```
matt-greathouse, line 133: The new `Fingerprint.start()` path only sets `startOptions.endpoints` to the single custom `behaviorUrl`, dropping the automatic fallback to the provider's default endpoint that the legacy `Fingerprint.load()` path still provides (which builds `[customUrl, defaultEndpoint]` for failover).

When `FINGERPRINT_JS_BEHAVIOR_URL` (custom subdomain/proxy) is configured together with this new v4 CDN loader path, `Fingerprint.start({endpoints: behaviorUrl})` has no secondary endpoint to fall back to if the custom subdomain is unreachable — a custom-subdomain outage now fails identification entirely instead of silently falling back to Fingerprint's default endpoint.

Consider passing an array (`[options.behaviorUrl, <default endpoint>]`) to `startOptions.endpoints` to preserve the failover behavior.

---

matt-greathouse (commented):
Automated code review — 3 findings, all in the bot-check/fingerprint logic:

1. **`pages/bot_check.py:854`** — missing fallback for `device_memory` produces a literal "None GB" in the report for non-Chromium browsers.
2. **`pages/bot_check.py:1215`** — a status-overwrite bug can mask a reported client-side fingerprint integration error (e.g. `csp_block`) with a generic `browser_only`/`unavailable` status.
3. **`static/js/account_identity_signals.js:133`** — the new `Fingerprint.start()` path drops the default-endpoint failover that the legacy `.load()` path preserves, so a custom-subdomain outage now fails identification outright.

No blocking security issues found. See inline comments for details and suggested fixes.

---

PR #1467, opened 2026-07-29
Improve Discord server access authorization

matt-greathouse:
## Summary
- Add explicit Discord guild authorization and migration support
- Update bot, API, settings, and legacy cleanup behavior
- Refresh native Discord app controls, agent UI, and documentation
- Add focused Discord bot unit test coverage

## Testing
- Not run (not requested)

### DIFF
--- a/api/services/discord_bot.py
+++ b/api/services/discord_bot.py
@@ -63,12 +63,10 @@ DISCORD_API_BASE = "https://discord.com/api/v10"
 DISCORD_OAUTH_AUTHORIZE_URL = "https://discord.com/oauth2/authorize"
 DISCORD_OAUTH_TOKEN_URL = "https://discord.com/api/oauth2/token"
 DISCORD_WEBHOOK_USERNAME_MAX_LENGTH = 80
-DISCORD_MANAGE_GUILD_PERMISSION = 0x20
-DISCORD_ADMINISTRATOR_PERMISSION = 0x8
 DISCORD_TEXT_CHANNEL_TYPES = {0, 5}
 DISCORD_WEBHOOK_MAX_FILES = 10
-DISCORD_OAUTH_USER_SCOPES = ("identify", "guilds")
 DISCORD_OAUTH_BOT_INSTALL_SCOPES = ("bot", "applications.commands")
+DISCORD_GUILD_INSTALL_TYPE = 0
 
 
 class DiscordBotIntegrationError(RuntimeError):
@@ -117,7 +115,21 @@ def _agent_owner(agent: PersistentAgent) -> tuple[Any, Any]:
 
 def _claimed_guild_queryset(agent: PersistentAgent):
     owner_user, organization = _agent_owner(agent)
+    return claimed_guild_queryset_for_owner(
+        owner_user=owner_user,
+        organization=organization,
+    )
+
+
+def claimed_guild_queryset_for_owner(*, owner_user=None, organization=None, include_legacy: bool = False):
+    if (owner_user is None) == (organization is None):
+        raise ValueError("Exactly one Discord owner must be provided.")
+
     queryset = PersistentAgentDiscordGuild.objects.filter(is_active=True)
+    if not include_legacy:
+        queryset = queryset.filter(
+            authorization_source=PersistentAgentDiscordGuild.AuthorizationSource.EXPLICIT_OAUTH,
+        )
     if organization is not None:
         return queryset.filter(organization=organization)
     return queryset.filter(owner_user=owner_user)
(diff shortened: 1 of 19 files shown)


api/services/discord_bot.py
```diff
@@ -433,95 +452,171 @@ def handle_discord_oauth_callback(
...
+            "last_synced_at": timezone.now(),
+        }
+        guild_claim = _claim_discord_guild_for_session(
+            session,
+            guild_id=authoritative_guild_id,
+            defaults=defaults,
+        )
+        if guild_claim is None:
+            raise DiscordBotIntegrationError(
+                "This Discord server is already connected to another Gobii context."
```
matt-greathouse, line 476: **Orphaned bot install when the guild is already claimed elsewhere.**

Because the Discord app uses Guild Install + `bot` scope, Discord adds the Gobii bot to the target server as soon as the user approves the OAuth consent screen — before this callback even runs (per the comment added in `config/settings.py`: "the token exchange identifies the one server selected during authorization"). So by the time `_claim_discord_guild_for_session` returns `None` here (guild already actively claimed by a different owner/org), the bot has *already* joined the Discord server.

This branch raises `DiscordBotIntegrationError` but never calls `_leave_discord_guild(authoritative_guild_id)` (or removes webhooks) to undo that install. The result: the rejected user sees an error, but the Gobii bot is left as a permanent, untracked member of that Discord server — and since they don't own the active claim, they have no way to remove it themselves (only an admin running `cleanup_legacy_discord_guilds` or similar could clean it up, and this guild wouldn't even match that command's `LEGACY_DISCOVERED` filter).

Consider calling `_leave_discord_guild(authoritative_guild_id)` in this branch before raising, mirroring the cleanup done in `disconnect_discord_guild_claim`.

---

api/services/discord_bot.py
```diff
@@ -209,99 +219,113 @@ def _oauth_redirect_uri() -> str:
...
+            f"{DISCORD_API_BASE}/guilds/{guild_id}",
+            headers=_discord_bot_headers(),
+            timeout=20,
+        )
+    except requests.RequestException as exc:
+        raise DiscordBotIntegrationError(
+            "Gobii could not verify the installed Discord server."
+        ) from exc
+    _raise_for_discord_status(response, action="installed server verification")
+    payload = response.json() or []
```
matt-greathouse: Nit: `payload = response.json() or []` looks like a copy-paste leftover from the old list-returning `_fetch_oauth_guilds` this replaced. `_fetch_bot_guild` returns a single guild object, so a falsy JSON response should fall back to `{}`, not `[]`. Harmless today (the `isinstance(payload, Mapping)` check right below still catches it and raises the correct error), but worth fixing for clarity — same pattern reused correctly for the list case in `_fetch_bot_channels` further down.

---

frontend/src/components/mcp/DiscordNativeAppModal.tsx
```diff
@@ -210,8 +248,14 @@ export function DiscordConfigurationScreen({
               guild={guild}
               subscriptions={app.subscriptions}
               selectedSubscriptions={selectedSubscriptions}
-              disabled={disabled}
+              disabled={busy}
+              isRemoving={pendingGuildId === guild.guildId}
               onToggleChannel={toggleChannel}
+              onRemove={() => {
```
matt-greathouse, line 254: This new "Remove server" flow (confirm dialog → `guildRemoval.mutate` → DELETE `/console/api/discord/guilds/<id>/`) is a destructive, context-wide action — it disables every agent's subscriptions in that server and uninstalls the bot — but it isn't exercised by any frontend test. `AgentPipedreamAppsModal.test.tsx` only adds `disconnectDiscordGuild: vi.fn()` to the mock module, without ever clicking "Remove" or asserting the success/error/query-invalidation behavior. Given the blast radius, a rendered test for this button (success path, and the 502/error path surfaced via `onError`) would be worth adding.

---

matt-greathouse (commented):
Automated review of the Discord explicit-guild-authorization changes.

Overall this is a solid, well-tested rework: moving from a broad "discover all manageable guilds" OAuth flow to a single-guild Guild Install flow closes a real over-authorization gap, and the new `authorization_source` migration/backfill, per-guild disconnect endpoint, and legacy-cleanup management command are all sensible additions with good test coverage (499 new/changed lines in `tests/unit/test_discord_bot.py` covering the conflict, migration, and partial-failure paths).

Three things worth a look before merge — one real correctness gap, one cosmetic nit, and one test-coverage gap:

1. **`api/services/discord_bot.py` (guild-claim conflict path)** — when OAuth completes for a guild that's already actively claimed by a different owner/org, the code raises an error without leaving the Discord guild, which the app's Guild Install flow will have already joined by that point. That looks like it leaves an orphaned, untracked bot membership with no way for the rejecting user to clean it up.
2. **`api/services/discord_bot.py:325`** — `_fetch_bot_guild`'s empty-payload fallback uses `[]` instead of `{}` (harmless today, just confusing).
3. **`DiscordNativeAppModal.tsx`** — the new "Remove server" UI action (disable-all-subscriptions + bot uninstall) has no frontend test coverage of the click → confirm → mutate → success/error flow.

No blocking issues otherwise — nice work on the transaction/HTTP-call separation and the partial-failure handling in `disconnect_discord_native_integration`.

---

PR #1469, opened 2026-07-29
Add asynchronous MCP task support

matt-greathouse:
## Summary

- Add persistent asynchronous MCP task tracking and background execution.
- Extend MCP manager and task protocol support for async tool workflows.
- Add agent follow-up handling, periodic task processing, and browser task integration.
- Update configuration, migrations, CI registration, and related unit tests.

## Testing

- Added and updated targeted unit tests for asynchronous MCP tasks, MCP tools, and browser task follow-ups.
- Not run (not requested).


api/models.py
```diff
@@ -13770,6 +13848,30 @@ def refresh_mcp_tool_cache_for_server(sender, instance, **kwargs):
...
+    if not instance.pk or instance.is_active:
+        return
+    was_active = MCPServerConfig.objects.filter(pk=instance.pk, is_active=True).exists()
+    if was_active:
+        from api.tasks.mcp_tasks import cancel_active_mcp_tasks_for_server
+
+        cancel_active_mcp_tasks_for_server(
+            str(instance.pk),
+            reason="The MCP server was disabled.",
+        )
```
matt-greathouse, line 13865: **Blocking network calls inside a `pre_save`/`pre_delete` signal, executed within an open DB transaction on the request thread.**

`cancel_active_mcp_tasks_for_server` (called synchronously here, and again from `cancel_mcp_tasks_for_deleted_server` below) iterates every active `PersistentAgentMCPTask` for the server and, via `_finish_task` → `_cancel_remote_task` → `MCPToolManager.cancel_mcp_task_remote`, makes a blocking `httpx` call (up to ~10s timeout) per task to the remote MCP server.

Both signals fire from request-handling views that already hold an open transaction, e.g. `console/api_views.py`:
- `MCPServerDetailAPIView.patch` wraps `form.save()` in `with transaction.atomic():` (this triggers `cancel_mcp_tasks_for_disabled_server` when `is_active` flips to `False`)
- `MCPServerDetailAPIView.delete` calls `server.delete()` directly, and Django's deletion `Collector` wraps the whole delete in its own `transaction.atomic()`

So disabling/deleting a server with several active outbound tasks (or a slow/unreachable remote MCP endpoint) serially blocks the request thread for potentially tens of seconds while holding the DB transaction/connection open — risking connection-pool exhaustion and request timeouts, and it's remotely triggerable by anyone who can rack up active tasks on a server before disabling/deleting it.

Suggest deferring the cancellation sweep to a Celery task dispatched via `transaction.on_commit(...)` rather than calling `cancel_active_mcp_tasks_for_server` inline in the signal handler.

---

api/tasks/mcp_tasks.py
```diff
@@ -0,0 +1,499 @@
...
+    else:
+        record_mcp_task_lifecycle("polled", task, remote_status=remote.status)
+        _apply_remote_state(task_id, lease_token, remote)
+        return
+
+    with transaction.atomic():
+        locked = PersistentAgentMCPTask.objects.select_for_update().get(pk=task_id)
+        if locked.terminal_at is None and locked.lease_token == lease_token:
+            _terminalize_locked(
+                locked,
```
matt-greathouse, line 393: Every code path that can fail while polling (timeouts, `MCPTaskHTTPError`, other `httpx.HTTPError`/`MCPTaskProtocolError`) is caught and handled here, but a genuinely unexpected exception from `get_mcp_manager().get_mcp_task_state(task)` (e.g. a bug in an adapter/normalizer) would propagate out of the Celery task uncaught. Since `_claim_task` doesn't touch `attempts`/`next_poll_at`, once `lease_expires_at` passes (`MCP_ASYNC_TASK_LEASE_SECONDS`, default 120s) `reconcile_mcp_tasks` will simply re-dispatch the same task, and it can hot-loop on the same bug indefinitely without the exponential backoff that `_schedule_retry` gives the handled-error paths.

Worth considering a catch-all fallback (mark the task `FAILED` or at least bump `attempts`) so a persistent bug degrades gracefully instead of spinning every lease interval.

---

api/agent/core/prompt_context.py
```diff
@@ -4761,6 +4811,13 @@ def _get_unified_history_prompt(
                 and _extract_spawn_web_task_task_id(result_text) in completed_browser_task_ids
             ):
                 continue
+            pending_mcp_task_id = _extract_spawn_web_task_task_id(result_text)
```
matt-greathouse, line 4814: Nit: `_extract_spawn_web_task_task_id` is reused here to pull the `task_id` out of MCP task creation responses. It works because both browser-task and MCP-task "pending" payloads happen to use the same `task_id` key, but the browser-task-specific name is misleading at this call site. Consider renaming to something generic like `_extract_task_id_from_result` since it's now shared by two unrelated task kinds.

---

matt-greathouse (commented):
Reviewed the async MCP task support (persistent `PersistentAgentMCPTask` model + migration, `mcp_task_protocol.py` stateless HTTP client, `mcp_manager.py` task creation/polling/cancellation, Celery polling/reconciliation in `mcp_tasks.py`, and the prompt-context wiring). Overall the design is solid: the lease-based claim in `_claim_task`, optimistic-concurrency guards around `wake_enqueued_at`, the minute-cadence reconciler for missed wakes/expiry/stuck leases, and the test coverage across the polling state machine are all well thought through.

One correctness/reliability concern worth addressing before merge (see inline comment on `api/models.py`): the `pre_save`/`pre_delete` signal handlers that cancel outstanding MCP tasks make blocking outbound HTTP calls synchronously, inside a DB transaction that's already open on the request thread (confirmed via `console/api_views.py`'s `MCPServerDetailAPIView.patch`/`.delete`). This can turn a simple "disable/delete MCP server" request into a multi-second-or-worse blocking call per active task.

Also left two smaller suggestions: a catch-all fallback for unexpected exceptions in `poll_mcp_task` so a code bug can't hot-loop every lease interval without backoff, and a naming nit around reusing a browser-task-specific helper for MCP task IDs.

---

ai-christianson:
Production Yeet is currently blocked because main contains two concurrent `api` migration leaves named `0442`. I opened hotfix #1470 adding the standard no-op merge migration as `0443_merge_20260729_1737`. Please rebase after #1470 lands and make this PR's current `0443_persistentagentmcptask` follow the merge node (normally renumber it to `0444` and depend on `0443_merge_20260729_1737`). This avoids reintroducing a second migration leaf.

PR #1472, opened 2026-07-29
Add MCP-native agent attachments

ai-christianson:
## Outcome

Makes Gobii MCP attachments usable in both directions without adding a parallel storage system:

- file and timeline results expose scoped `resource_link` blocks and `gobii://` URIs
- standard `resources/read` returns text/blob resource contents
- `gobii_read_agent_resource` gives tool-centric clients native image/resource content
- `gobii_send_agent_message` can upload and attach up to 10 small files in one call
- every read rechecks agent access, enforces the platform size limit, and verifies stored checksums

The implementation follows the stable MCP 2025-11-25 resource primitives and keeps existing upload-then-send clients compatible.

## Red -> green evidence

The new protocol tests initially failed because Gobii advertised no resource capability, returned no resource URIs, rejected `resources/read`, and rejected inline message attachments. They now cover:

- filespace image resource links, direct reads, and native image content
- unimported timeline attachment reads
- cross-agent authorization isolation
- checksum mismatch rejection
- validation of all inline files before any write
- one-call upload + message attachment with text resource readback
- resource/tool discovery

## Local verification

- `RemoteMCPViewTests`: 37/37 green
- focused attachment/timeline tests: green
- Python compile check: green
- complexity LoC guard: green
- `git diff --check`: green

No behavioral prompt eval was added: this is deterministic MCP protocol, authorization, storage, and serialization behavior, fully covered at the real HTTP endpoint boundary.

### DIFF
--- a/api/mcp_views.py
+++ b/api/mcp_views.py
@@ -11,7 +11,18 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from api.auth import MCPAPIKeyAuthentication
-from api.services.remote_mcp import MCPToolError, MCP_PROTOCOL_VERSION, SERVER_INFO, call_tool, list_tools, make_tool_result
+from api.services.remote_mcp import (
+    MCPToolError,
+    MCP_PROTOCOL_VERSION,
+    SERVER_INFO,
+    call_tool,
+    list_resource_templates,
+    list_resources,
+    list_tools,
+    make_tool_result,
+    read_resource,
+)
+from api.services.remote_mcp_resources import MCPResourceError
 
 
 JSON_RPC_PARSE_ERROR = -32700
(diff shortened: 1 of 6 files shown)


ai-christianson:
Exact-head preview verification passed on `9c26aee21` at `https://pr-1472.ship.gobii.ai/api/v1/mcp/`.

Live MCP round trip:
- initialized and advertised `resources`
- discovered the inline `attachments` send field and `gobii_read_agent_resource`
- created an ephemeral preview agent
- sent a real PNG inline without processing
- verified returned `resource_link`, MIME, SHA-256, filespace URI, and timeline URI
- read identical bytes through `resources/read`
- received identical bytes as native MCP `image` content through the tool
- archived the QA agent

All checks passed. Full CI and docs are green. A read-only merge-tree check against Matt’s current async MCP PR #1469 head (`9dd45a867`) also merged cleanly, so the two MCP changes can coexist without a source conflict.

PR #1482, opened 2026-07-30
Improve immersive app performance and agent chat state handling

matt-greathouse:
## Summary

- Improve immersive app performance through optimized agent roster, timeline, and shell state handling
- Refine agent chat settings, context switching, composer, sidebar, and embedded panels
- Add cache refresh and profile synchronization improvements
- Expand coverage for agent chat access, APIs, roster behavior, setup flows, and console functionality

## Testing

- Not run (not requested)

### DIFF
--- a/console/api_views.py
+++ b/console/api_views.py
@@ -3322,11 +3322,16 @@ class AgentProfileAPIView(LoginRequiredMixin, View):
     http_method_names = ["get"]
 
     def get(self, request: HttpRequest, agent_id: str, *args: Any, **kwargs: Any):
-        agent = resolve_agent_for_request(
-            request,
-            agent_id,
-            allow_shared=True,
-            allow_delinquent_personal_chat=True,
+        staff_override = get_staff_context_override(request)
+        agent = (
+            resolve_staff_agent(request.user, agent_id, staff_override)
+            if staff_override
+            else resolve_agent_for_request(
+                request,
+                agent_id,
+                allow_shared=True,
+                allow_delinquent_personal_chat=True,
+            )
         )
         return JsonResponse(
             _serialize_agent_payload(
--- a/console/api_views.py
+++ b/console/api_views.py
@@ -2858,7 +2858,7 @@ def _web_chat_properties(agent: PersistentAgent, extra: dict[str, Any] | None = 
     return Analytics.with_org_properties(payload, organization=getattr(agent, "organization", None))
 
 
-def _serialize_agent_profile_payload(
+def _serialize_agent_payload(
     request: HttpRequest,
     agent: PersistentAgent,
     *,
@@ -2872,10 +2872,13 @@ def _serialize_agent_profile_payload(
     is_admin_user: bool | None = None,
     can_reactivate_agent: bool | None = None,
     enrich: bool = False,
+    include_profile_details: bool = True,
 ) -> dict[str, Any]:
     user = request.user
     if enrich:
         enrich_agents_for_card_surface([agent], owner or agent.organization or agent.user)
+    elif not include_profile_details:
+        agent.display_tags = agent.tags if isinstance(agent.tags, list) else []
     if is_collaborator is None:
         is_collaborator = user_is_collaborator(user, agent)
     if processing_active is None:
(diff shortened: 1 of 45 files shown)


frontend/src/screens/AgentChatPage.tsx
```diff
@@ -2027,11 +1988,14 @@ export function AgentChatPage({
...
-  const activeAgentIsActive = activeRosterMeta?.isActive ?? activeChatSession.identity.agentIsActive
+  const activeAgentIsActive = activeChatSession.identity.agentIsActive === false
+    ? false
+    : activeRosterMeta?.isActive ?? true
   const activeCanReactivateAgent = activeRosterMeta?.canReactivateAgent
     ?? activeChatSession.identity.canReactivateAgent
-  const activeCanSendMessages = activeRosterMeta?.canSendMessages ?? !staffContext
-  const activeCanManageCollaborators = activeRosterMeta?.canManageCollaborators ?? (canManageCollaborators ?? true)
+  const activeProfileReady = isNewAgent || activeProfileQuery.isSuccess
+  const activeCanSendMessages = activeProfileReady && Boolean(activeRosterMeta?.canSendMessages)
```
matt-greathouse, line 1997: `activeCanSendMessages` now requires `activeProfileReady` (i.e. `activeProfileQuery.isSuccess`), whereas before it defaulted to `!staffContext` (optimistically `true` for normal users). Since `activeProfileQuery` is a separate, non-prefetched network round trip keyed off `activeAgentId` with no `placeholderData`/`initialData`, every time a user opens an agent whose profile isn't already cached (new tab, first visit, or after the 5‑minute `staleTime` expires), `activeCanSendMessages` will be `false` until that fetch resolves.

That feeds directly into `sendMessageDisabledReason` a few hundred lines down:
```ts
const sendMessageDisabledReason = !isNewAgent && !activeCanSendMessages
  ? 'This staff view is read-only. You can send a system message in Developer Mode.'
  : ...
```
So regular (non-staff) users will briefly see the composer disabled with a "This staff view is read-only" message on essentially every fresh agent open, since `initialLoading` (which gates the composer's visibility) only depends on `timelineQuery`, not on `activeProfileQuery` — the two requests aren't synchronized. Worth gating this message on `staffContext` explicitly, or giving `activeProfileQuery` `placeholderData`/an optimistic default so the fail-closed behavior doesn't surface a staff-specific message to normal users.

---

frontend/src/components/agentChat/ChatSidebarGallery.tsx
```diff
@@ -136,12 +138,12 @@ function GalleryCard({
               <Settings className="h-3.5 w-3.5" />
               <span>Configure</span>
             </button>
-          ) : agent.detailUrl ? (
-            <a className="agent-gallery-card__primary-action" href={agent.detailUrl}>
+          ) : (
+            <a className="agent-gallery-card__primary-action" href={`/app/agents/${agent.id}/settings`}>
```
matt-greathouse: Now that `detail_url` is dropped from the (lightweight) roster payload, this hardcodes `/app/agents/${agent.id}/settings` instead of using the server-computed `detailUrl`. That loses the special case in `serialize_agent_card_payload` (`console/agent_cards.py`), where `detail_url` is set to `chat_url` instead of the settings URL when `is_shared` is true. If a user who is both an explicit `AgentCollaborator` and otherwise able to manage the agent (e.g. an org member who was also individually added as a collaborator) sees this card, "Configure" now always routes to the settings page rather than chat, which is a behavior change from before this PR. Worth confirming this is intentional, or keeping `isCollaborator` in the roster payload and branching on it client-side to preserve the old routing.

---

console/api_views.py
```diff
@@ -7287,6 +7281,7 @@ def post(self, request: HttpRequest, agent_id: str, *args: Any, **kwargs: Any):
         agent, owner, plan_payload, can_manage_billing, can_open_billing = self._resolve_agent_addons_context(
             request,
             agent_id,
+            reconcile_personal_plan=True,
```
matt-greathouse, line 7289: `reconcile_personal_plan=True` is only passed here, in `post()`. `get()` (line 7271) calls `_resolve_agent_addons_context` with the default `reconcile_personal_plan=False`, so it now resolves the personal plan via `get_user_plan` (local/cached, no Stripe sync) instead of `reconcile_user_plan_from_stripe`. That means a user landing on the addons/billing panel right after completing a Stripe checkout (before the webhook lands) will see stale plan/contact-pack data on the `GET`, and it only reconciles once they perform a mutating action. Likely intentional for this PR's perf goal (avoiding a live Stripe call on every panel load), but worth double-checking there's no post-checkout flow relying on `GET` reflecting the new plan immediately.

---

matt-greathouse (commented):
Reviewed the diff (40 files, +537/-420) via `/review`. Overall this is a solid, well-tested performance pass: the roster/profile payload split (lightweight roster list vs. full profile fetched on demand) removes real N+1 prefetches (`system_skill_states`, `preferred_llm_tier`) and an unconditional Stripe reconciliation on nearly every agent-related request, backed by matching test updates. The React-side memoization (`memo()` on `GalleryCard`/`AgentListItem`), `useDeferredValue` on sidebar mode, lazy avatar loading, and `content-visibility` CSS hints are all reasonable, low-risk wins.

Left 3 comments on things worth a second look before merge:
1. **Likely UX regression**: `activeCanSendMessages` now gates on the new `agent-profile` query succeeding, and since that query isn't prefetched/synced with the timeline load, normal users can briefly see the composer disabled with a staff-specific "read-only" message on every fresh agent open.
2. **Behavior change**: `GalleryCard`'s "Configure" link now hardcodes the settings URL instead of using the server-computed `detailUrl`, which drops the `is_shared → chat_url` routing for collaborators in an edge case.
3. **Question**: the addons `GET` endpoint no longer reconciles the personal plan with Stripe (only `POST` does now) — please confirm no post-checkout flow depends on `GET` reflecting a fresh plan.

No security or test-coverage gaps beyond the above — the Python-side changes are well covered by updated unit tests.

---

PR #1484, opened 2026-07-30
Add computer C++ connection integration

matt-greathouse:
## Summary

- Add computer C++ connection and relay transport support across backend, frontend, and agent tooling
- Add connection management APIs, UI panels, insights, system-skill registration, and persistence
- Add periodic relay tasks, migrations, eval scenarios, and CI registration

## Testing

- Added focused unit and eval coverage for the computer C++ integration
- Not run (not requested)


console/computer_api.py
```diff
@@ -0,0 +1,280 @@
...
+            raise PermissionDenied("Computer connections are not enabled for this account")
+        device = _owner_device(request, device_id)
+        try:
+            payload = parse_computer_json_payload(request)
+            display_name = None
+            if "display_name" in payload:
+                display_name = str(payload["display_name"]).strip()[:128]
+                if not display_name:
+                    raise ValueError("display_name cannot be empty")
+            paused = bool(payload["paused"]) if "paused" in payload else None
```
matt-greathouse: `bool(payload["paused"])` coerces via Python truthiness rather than validating an actual boolean. A client sending `PATCH {"paused": "false"}` (a non-empty string) gets `bool("false") == True`, silently pausing the device instead of unpausing it — the opposite of the caller's intent, and it also fires `relay.state` to close the relay for legitimate use. Suggest explicitly checking `payload["paused"] is True/False` or using a strict bool parser and rejecting other types with a 400.

---

console/computer_api.py
```diff
@@ -0,0 +1,280 @@
...
+                if not isinstance(payload["approved_apps"], list):
+                    raise ValueError("approved_apps must be a list")
+                approve_device_apps(device, payload["approved_apps"])
+            if payload.get("agent_id"):
+                agent = get_object_or_404(PersistentAgent, id=payload["agent_id"])
+                assign_device(device, agent, granted_by=request.user)
+        except ValueError as exc:
+            return HttpResponseBadRequest(str(exc))
+        except PermissionError as exc:
+            raise PermissionDenied(str(exc)) from exc
```
matt-greathouse, line 249: `update_device_properties()`, `approve_device_apps()`, and `assign_device()` each open their own internal `transaction.atomic()` in the service layer, but this view doesn't wrap all three in a single outer transaction. If a later step raises (e.g. `assign_device()` raises `PermissionError` because the caller can't manage the target agent), earlier steps' writes — like a `display_name` rename — are already committed even though the overall request returns 4xx. Consider wrapping the whole `try` block in `transaction.atomic()` so a failed PATCH is fully rolled back.

---

api/agent/tools/computer_relay_transport.py
```diff
@@ -0,0 +1,88 @@
...
+                        mode="json",
+                        by_alias=True,
+                        exclude_none=True,
+                    )
+                    try:
+                        response_payload = await relay_mcp_request(
+                            self.device_app_id,
+                            request_payload,
+                        )
+                        if not response_payload:
```
matt-greathouse: `if not response_payload: continue` treats a valid-but-empty `{}` response the same as "nothing to forward" and silently drops it — but `relay_mcp_request` (`api/services/computer_relay.py`) explicitly treats `{}` as a valid successful dict response. If the native client ever replies with an empty-but-valid payload, the blocked `ClientSession` request on `client_receive` never gets a correlated response and hangs until an external timeout fires, instead of failing fast the way a malformed non-dict payload does a few lines below. Consider checking `response_payload is None` instead of falsy.

---

api/computer_consumers.py
```diff
@@ -0,0 +1,254 @@
...
+            await self.send_json(
+                {"type": "error", "error": {"code": "invalid_message", "message": "Unknown message type"}}
+            )
+
+    async def _reject(self, code: str, message: str, *, close_code: int):
+        await self.send_json({"type": "error", "error": {"code": code, "message": message}})
+        await self.close(code=close_code)
+
+    async def _handle_hello(self, content):
+        if int(content.get("protocol_version") or 0) != settings.COMPUTER_CPP_RELAY_PROTOCOL_VERSION:
```
matt-greathouse: `int(content.get("protocol_version") or 0)` isn't guarded against `ValueError`/`TypeError`, unlike the equivalent cast in `create_pairing_session` (`api/services/computer_relay.py`), which wraps the same parse in `try/except (TypeError, ValueError)`. A connected device sending `{"type": "hello", "protocol_version": "abc", ...}` raises here uncaught, propagating out of `receive_json` instead of closing gracefully with an `error` frame + close code like every other malformed-input path in this consumer. Depending on how Channels handles an exception raised mid-`receive`, `disconnect()` may not run cleanly, leaving stale Redis presence for the device until `COMPUTER_CPP_PRESENCE_TTL_SECONDS` expires.

---

api/computer_consumers.py
```diff
@@ -0,0 +1,254 @@
...
+            await self.channel_layer.send(
+                previous["channel"],
+                {
+                    "type": "computer.control",
+                    "event_type": "relay.close",
+                    "payload": {"code": "superseded"},
+                },
+            )
+            clear_device_presence(self.device.id, previous["generation"])
+        self.socket_counted = True
```
matt-greathouse, line 64: `connect()` has no handshake timeout, and the "supersede the previous connection" check only detects a prior connection via `get_device_presence`, which is only populated once `_handle_hello` succeeds (line ~160). A socket that authenticates but stalls before sending `hello` is invisible to this check. Scenario: device connects (socket A, valid token) but stalls before `hello` (flaky network); the client retries with a fresh token on a second connection (socket B) — since A never reached `_handle_hello`, presence is still empty, so B never sends A a `superseded` close. If B completes the handshake, A remains a live, half-open, uncounted socket indefinitely — a per-device resource leak and a bypass of the intended single-active-connection invariant. Consider closing any not-yet-helloed socket after a short timeout.

---

api/models.py
```diff
@@ -8063,6 +8097,243 @@ def headers(self, value: dict[str, str] | None) -> None:
...
+        APPROVED = "approved", "Approved"
+        DENIED = "denied", "Denied"
+        REDEEMED = "redeemed", "Redeemed"
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    device_code_digest = models.CharField(max_length=64, unique=True)
+    user_code_digest = models.CharField(max_length=64)
+    machine_identifier_digest = models.CharField(max_length=64)
+    display_name = models.CharField(max_length=128)
+    platform = models.CharField(max_length=32)
```
matt-greathouse: `ComputerPairingSession.platform` has no `choices` constraint, unlike the near-identical `ComputerDevice.platform` field which restricts values to `Platform.choices` (macos/windows). The macos/windows restriction currently lives only in `create_pairing_session`'s manual `if platform not in ComputerDevice.Platform.values: raise ValueError` — any other write path (fixture, admin script, future API variant) bypasses that check, and since Django `choices` aren't a DB constraint, an arbitrary string could get persisted here and then copied straight into a new `ComputerDevice` row by `redeem_pairing()`. Consider adding `choices=ComputerDevice.Platform.choices` here too for defense in depth.

---

api/models.py
```diff
@@ -8063,6 +8097,243 @@ def headers(self, value: dict[str, str] | None) -> None:
...
+    byte_count = models.PositiveIntegerField()
+    sha256 = models.CharField(max_length=64)
+    expires_at = models.DateTimeField(db_index=True)
+    consumed_at = models.DateTimeField(null=True, blank=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        indexes = [
+            models.Index(fields=["device", "expires_at"], name="computer_artifact_exp_idx"),
+        ]
```
matt-greathouse: `ComputerRelayArtifact` stores screenshots in `default_storage` via `storage_key`, but there's no `post_delete` signal (or `delete()` override) to remove the backing blob when rows are deleted other than through the explicit `revoke_device()`/`consume_artifact()` service calls that manually call `default_storage.delete()` first. Since `device` is `on_delete=CASCADE` from `ComputerDevice`, which itself cascades from `User`, any hard-delete of a user (account deletion/GDPR erasure) or a future bulk-cleanup job (`ComputerRelayArtifact.objects.filter(expires_at__lt=now).delete()`) removes rows via Django's CASCADE machinery without ever touching object storage, leaking the image blobs permanently. This codebase already has the right pattern for this exact case — `delete_user_pet_spritesheet`, a `post_delete` signal using `transaction.on_commit` (fires even for cascade-collected deletes, unlike an overridden instance `delete()`) — worth mirroring here.

---

api/agent/system_skills/defaults.py
```diff
@@ -1192,6 +1252,49 @@ def _hubspot_native_prompt_instructions(agent) -> str:
...
+    discovery_triggers=(
+        "computer",
+        "desktop",
+        "local machine",
+        "screen",
+        "keyboard",
+        "mouse",
+        "windows",
+        "mac",
+    ),
```
matt-greathouse, line 1285: This skill definition never sets `prompt_available` (unlike, e.g., the custom-tool-development skill's `_custom_tool_development_prompt_available`), so `should_render_prompt()` returns `True` unconditionally — the discovery suggestion is surfaced to every agent regardless of whether the owning account has the `computer_cpp_integration` waffle flag enabled. Combined with very common `discovery_triggers` ("computer", "screen", "windows", "mac"), an agent belonging to a flagless user whose recent message contains e.g. "my computer crashed" gets a "REQUIRED NEXT RELEVANT ACTION: enable Computer skill" injected into its prompt, telling the user to go install computer.cpp — a feature they can't actually use since pairing is separately gated by `computer_cpp_enabled_for_user()`. Consider adding a `prompt_available` callback that checks the flag before this skill is suggested.

---

api/agent/system_skills/defaults.py
```diff
@@ -44,6 +45,65 @@ def _app_integrations_url() -> str:
...
+            state = "paused"
+        elif get_device_presence(device.id):
+            state = "online"
+        else:
+            state = "offline"
+        approved_apps = [
+            app.display_name
+            for app in device.apps.all()
+            if app.approval_state == app.ApprovalState.APPROVED and app.is_available
+        ]
```
matt-greathouse, line 93: This lists an app as "approved" based only on `approval_state == APPROVED and is_available`, but doesn't check `approved_schema_hash == reported_schema_hash` the way `_reconcile_app_configs` (`api/services/computer_relay.py`) does before it will actually keep the corresponding `MCPServerConfig` active. If the desktop app's manifest schema changes (e.g. a computer.cpp version bump), `_reconcile_app_configs` disables the underlying tool registration pending re-approval, but this prompt-context renderer still reports the app as approved — the agent will believe an `mcp_computer_...` tool exists, try to call it, and get a confusing "Unknown MCP tool" error instead of being told the app needs re-approval.

---

frontend/src/components/mcp/ComputerConnectionsPanel.tsx
```diff
@@ -0,0 +1,470 @@
...
+                        </div>
+                      </td>
+                      <td className="px-6 py-4">
+                        <div className="flex justify-end gap-2">
+                          {device.permissions.can_manage_device ? (
+                            <>
+                              <SettingsActionButton
+                                surface={buttonSurface}
+                                size="sm"
+                                onClick={() => updateComputer(url, device.id, { paused: !device.paused }).then(() => refresh(device.paused ? 'Computer resumed.' : 'Computer paused.'))}
```
matt-greathouse: This Pause/Resume handler (and the Unassign/Revoke/Remove-grant handlers below at lines 402, 413, and 425) calls `.then(...)` with no `.catch()`/error handling and no pending/disabled state, unlike the `useMutation`-based modals elsewhere in this same directory (e.g. `AgentPipedreamAppsModal.tsx`, `WorkspaceAppsManager.tsx`) which surface failures via `onError`. If the request rejects (network error, expired session, or the device's assignment changing concurrently), the promise rejects unhandled: no error banner appears, the button stays enabled the whole time, and the user has no indication the action failed — e.g. they may believe a computer was paused/unassigned/revoked when it wasn't. Worth routing these four actions through the same `useMutation` pattern used by `PairingModal`/`ManageComponentModal` in this file.

---

api/agent/tools/eval_synthetic_tools.py
```diff
@@ -146,6 +146,26 @@
...
+            "required": ["target"],
+            "additionalProperties": False,
+        },
+        "system_skill_key": "computer",
+    },
+    "mcp_computer_lab_pc_gobii_desktop_take_screenshot": {
+        "description": "Take a screenshot on the connected computer named Lab PC using Gobii Desktop.",
+        "parameters": {"type": "object", "properties": {}, "additionalProperties": False},
+        "system_skill_key": "computer",
+    },
```
matt-greathouse, line 190: Unlike every other tool family in `EVAL_SYNTHETIC_TOOL_DEFINITIONS` (google_sheets, apollo_io, mcp_brightdata_*, eval_*), these new `mcp_computer_*` tools got no matching branch added to `get_eval_synthetic_tool_fallback_result` (further down in this file, outside this diff's hunks — the `elif`/`else` chain around what is currently line ~366), so an eval scenario without an explicit fixture falls through to the fully generic `else` returning `{"items": [], "match_count": 0}`. A real computer-tool call actually returns MCP content/image results or one of the relay's `status: action_required`/`error` shapes (offline, paused, permissions_required, etc. — see `_computer_error_response` in `mcp_manager.py`). An eval testing "agent handles computer.cpp offline gracefully" without a specific fixture gets a fallback shape that looks nothing like the real thing, making such eval results unreliable.

---

matt-greathouse (commented):
Reviewed the full diff (36 files, ~4.9k added lines) across the relay/transport core, HTTP + console API, data models/migrations, agent-tool & system-skill integration, evals/tests, and the frontend UI. Overall the security fundamentals are solid: relay access tokens are HMAC-signed with `hmac.compare_digest` comparisons, secret-bearing model fields are stored as digests (consistent with existing `TrialPromo`/`ApiKey` conventions), device/agent ownership is checked in the service layer (not just at the view), and I didn't find an IDOR or unauthenticated-access path into another tenant's device/connection.

Left 11 line comments on concrete issues found:

**Correctness bugs**
- `console/computer_api.py`: a `bool(payload["paused"])` truthy-coercion bug means `{"paused": "false"}` incorrectly pauses instead of unpausing.
- `console/computer_api.py`: the PATCH handler's three mutating calls aren't wrapped in one outer transaction, so a later failure (e.g. `assign_device` permission error) leaves earlier writes (like a rename) committed.
- `api/agent/tools/computer_relay_transport.py`: a valid-but-empty `{}` relay response is silently dropped instead of forwarded, hanging the waiting MCP request instead of failing fast.
- `api/computer_consumers.py`: an unhandled `ValueError`/`TypeError` in `_handle_hello`'s protocol-version parsing can crash the consumer ungracefully instead of closing gracefully like other malformed-input paths.
- `api/computer_consumers.py`: no handshake timeout — a socket that authenticates but stalls before `hello` isn't detected by the "supersede previous connection" check, risking a per-device half-open-socket leak.

**Data model**
- `api/models.py`: `ComputerPairingSession.platform` lacks the `choices` constraint that `ComputerDevice.platform` has.
- `api/models.py`: `ComputerRelayArtifact` has no `post_delete` cleanup signal for its storage blob, so cascade-deletes (user deletion, future bulk cleanup) leak files in object storage — this codebase already has the right pattern elsewhere (`delete_user_pet_spritesheet`).

**Feature-flag gating**
- `api/agent/system_skills/defaults.py`: the Computer skill's discovery is not gated behind `prompt_available`/the `computer_cpp_integration` flag, so very common trigger words ("computer", "screen", "mac") can suggest this feature to users who can't actually use it.
- `api/agent/system_skills/defaults.py`: `approved_apps` in the prompt context doesn't check schema-hash agreement, so it can report an app as approved when the underlying tool registration was actually disabled pending re-approval.

**Test/eval infra**
- `api/agent/tools/eval_synthetic_tools.py`: no fallback-result branch for `mcp_computer_*` tools, so unfixtured eval cases get a generic shape that doesn't resemble real relay responses.

**Frontend**
- `frontend/src/components/mcp/ComputerConnectionsPanel.tsx`: the Pause/Resume, Unassign, Revoke, and Remove-grant action buttons all fire-and-forget their mutation promises with no `.catch()`/error surfacing or pending state, unlike sibling modals in the same file that use `useMutation`.

**Test coverage gaps** (not filed as line comments since they're absence-of-code, but worth flagging): `tests/unit/test_computer_cpp_integration.py` has no coverage for the cross-agent pairing-authorization check, the wrong-device-code rejection path, WS connect-time auth failures (missing/invalid token), or the reconnection/"superseded" + stale-generation guard paths — exactly the areas the relay-core comments above touch. Separately, `api/evals/scenarios/computer_integration.py`'s cases all use mocked tool results via the synthetic `"eval"` tool_server, so this eval suite never actually exercises `computer_relay.relay_mcp_request`/`ComputerRelayConsumer`, despite being tagged as a real-harness native-integration scenario.

No issues found in `config/settings.py`, `.github/workflows/ci.yml`, `util/analytics.py`, or the migration-vs-model parity check.

---

PR #1500, opened 2026-08-02
Tighten agent execution under pressure

ai-christianson:
## What changed

- make high-load continuation instructions use the actual plan vocabulary (`todo` / `done`) and prefer direct enabled-tool execution over redundant discovery, repeat calls, or reply-only SQLite staging
- make compacted message attribution use one targeted SQLite query and add real-harness gates for bounded retrieval plus error-free first-pass delivery
- keep implied-send system instructions stable across requester identities for better prompt caching
- simplify link guidance while preserving structured comparison tables, and safely normalize the common unambiguous `$[link:https://...]` mistake to its already-trusted raw URL
- lower all representative prompt-size ceilings to the new measured values

## Measured impact

Compared with main, representative rendered prompts are smaller:

- first run: 9,254 → 9,223 fitted tokens; 67,723 → 67,656 bytes
- normal explicit send: 8,697 → 8,666 fitted tokens; 64,958 → 64,892 bytes
- web implied send: 8,873 → 8,804 fitted tokens; 65,598 → 65,445 bytes

## Verification

- DeepSeek V4 Flash pressure suite: 26/26 tasks (6 runs) on the exact final prompt; raw traces show no error recovery in the new gate and 0–1 SQLite calls for compacted attribution
- DeepSeek V4 Flash oversized owner-report regression: 16/16 tasks (4 runs), each one fetch then one structured, linked 4/4 delivery
- broader pressure validation before the final micro-compression: 52/52 tasks (12 runs) plus 40/40 advisory tasks (8 runs)
- focused Django tests: 36/36
- link-reference focused regression: pass; broader link-reference unit run previously passed 40 tests with one local-only native PDF dependency unavailable on macOS
- complexity budgets and `git diff --check`: pass

Per project guidance, I did not run the crash-prone global Django suite; CI remains the full integration gate.


ai-christianson:
Exact-head release gate (`ac0a447`) passed: CI is green across all 10 backend shards, frontend, sandbox, tag, and tightened complexity guards; the matching preview deployment succeeded. Authenticated preview QA passed API ping (200), MCP initialize/tool discovery, and headed-browser login/render checks. A locked-down disposable agent completed both live probes in one shot: a simple greeting used exactly one same-channel send with no plan/SQLite detour, and a supplied 3-company comparison used exactly one send, a correctly linked Markdown table, no retries, and no tool errors. The table and all three links rendered correctly in the browser. The probe was archived. Main remains unchanged from the PR base and the diff is clean.

PR #1504, opened 2026-08-03
Improve agent chat roster and avatar performance

matt-greathouse:
## Summary

- Optimize agent chat roster rendering with virtualization and timeline cache injection
- Add indexed visible-agent message roster lookups and improve message read-state handling
- Add background avatar thumbnail generation and caching
- Refine chat sidebar components and avatar presentation
- Expand backend and frontend test coverage for the performance changes

## Testing

Not run (not requested)


console/agent_audit/events.py
```diff
@@ -141,109 +151,318 @@ def _event(payload: dict, *, timestamp: datetime, kind: str, identifier, develop
...
-    ]
 
 
-def _message_events(agent, cursor, direction, limit, at, developer):
-    queryset = (
-        PersistentAgentMessage.objects
-        .filter(owner_agent=agent)
-        .select_related("from_endpoint", "to_endpoint", "conversation__peer_link", "peer_agent", "owner_agent")
-        .prefetch_related("attachments__filespace_node", "cc_endpoints", "bcc_endpoints")
+def _postgres_candidates(
```
matt-greathouse, line 255: This new hand-rolled raw-SQL path (UNION ALL across 6 tables + manual cursor-pagination predicates in `_postgres_cursor_clause`) only executes when `connection.vendor == "postgresql"`. The same is true for `_postgres_latest_visible_outbound_messages` in `api/agent/comms/message_reads.py`.

CI's `config.test_settings` forces `ENGINE: django.db.backends.sqlite3` (see `config/test_settings.py:48`), and `.github/workflows/ci.yml` runs the whole test matrix with `DJANGO_SETTINGS_MODULE: config.test_settings` — the `postgres` service container in that job is never actually used as Django's DB backend. So this entire code path (the one that will run in production) has zero automated coverage; only the `_portable_*` fallback gets exercised by `manage.py test`.

Given the complexity here (custom cursor comparison operators per direction/kind, `IS DISTINCT FROM`, JSONB operators, casting identifiers to `varchar`/`uuid` per branch), I'd want at least one test run against a real Postgres instance before merging — a bug here would only surface in prod. Worth adding a Postgres-backed CI lane (or manual verification) rather than relying solely on the SQLite-only suite.

---

frontend/src/screens/AgentChatPage.tsx
```diff
@@ -1536,7 +1536,6 @@ export function AgentChatPage({
     enabled: contextReady,
     context: effectiveContext,
     contextKey: rosterContextKey,
-    forAgentId: routeAgentId ?? undefined,
```
matt-greathouse, line 1539: Dropping `forAgentId: routeAgentId ?? undefined` from the primary roster query looks like it could reintroduce a previously-fixed bug. `useAgentRoster.ts` still carries this comment right above where `forAgentId` feeds into the query key:

> "The roster scoped to a specific agent comes back with that agent's own context, which can differ from the one the key describes. Leaving forAgentId out of the key filed that response under the wrong entry, and the caller then discarded it as a context mismatch."

There's a dedicated regression suite for this (`useAgentRosterScoping.test.tsx`), but it only tests the hook in isolation — `AgentChatPage.test.tsx` fully mocks `useAgentRoster` (see its `vi.mock('../hooks/useAgentRoster', ...)`), so nothing here would catch a regression at this call site. The backend `for_agent` param (`console/api_views.py` around line 3095, `_resolve_override_for_agent`) does real context-override/inclusion logic for the requested agent, so a directly-linked or cross-context agent could silently be missing from (or misfiled in the cache for) the roster now.

If `useConsoleContextSwitcher`'s own `forAgentId: contextLookupAgentId` (kept a few lines up) is now sufficient to make this redundant, worth calling that out explicitly (in a comment or the PR description) so it's clear this was intentional rather than dropped incidentally during the virtualization refactor.

---

frontend/src/components/agentChat/ChatSidebarGallery.tsx
```diff
@@ -192,41 +199,58 @@ export function ChatSidebarGallery({
...
   const favoriteAgents = agents.filter((agent) => favoriteAgentIdSet.has(agent.id))
   const allAgents = agents.filter((agent) => !favoriteAgentIdSet.has(agent.id))
   const showFavoritesSection = favoriteAgents.length > 0
   const showAllSection = allAgents.length > 0 || !showFavoritesSection
+  const columnCount = viewportWidth > 0
+    ? variant === 'drawer' && viewportWidth < 560
+      ? 1
+      : Math.max(1, Math.floor((viewportWidth + 12) / (232 + 12)))
+    : 1
+  const rows: VirtualRosterRow[] = []
```
matt-greathouse: Unlike `ChatSidebar.tsx`'s equivalent row-building logic (wrapped in `useMemo`), this `rows` array — and everything pushed into it below — is rebuilt from scratch on every render of `ChatSidebarGallery`.

`VirtualizedRosterSurface`'s scroll-to-agent effect depends on the `rows` reference (`[enabled, onScrolledToAgent, rows, scrollToAgentId, virtualizer]`), so every re-render of the gallery (e.g. each tick of `refetchIntervalMs: ROSTER_REFRESH_INTERVAL_MS` in the roster query) will produce a new `rows` array and re-run that effect — potentially re-triggering `scrollToIndex`/`scrollIntoView` even when nothing scroll-relevant actually changed. It also partly offsets the perf goal of this PR, since all the row `content` JSX (including `GalleryCard`s) gets rebuilt on every render regardless of whether `agents`/`favoriteAgentIds`/etc. changed.

Worth memoizing this the same way `ChatSidebar` does (e.g. `useMemo` keyed on `agents`, `favoriteAgentIds`, `columnCount`, etc.).

---

matt-greathouse (commented):
Reviewed the diff directly (fetched `origin/main` and `origin/performance-improvements-8-3-26` and diffed locally, since `pull_request_read.get_diff` exceeded the tool's output limit).

**Overview:** Solid, well-scoped performance PR — roster list virtualization via `@tanstack/react-virtual`, an indexed/raw-SQL fast path for the "latest visible message" lookup and the audit timeline event merge (with a portable SQLite fallback kept for tests), lazy avatar image loading, and a new background Celery task to pre-warm avatar thumbnails. Backend permission-flag refactors (`can_manage_agent`, `can_send_messages`, `manageable_agent_ids`) were traced back to the original per-agent helper functions (`user_can_manage_agent_settings`, `user_has_natural_agent_chat_access`) and look behaviorally equivalent, just computed in bulk instead of N queries — nice, and backed by new query-count regression tests (`test_roster_query_count_is_bounded_as_agents_grow`, `test_developer_timeline_message_queries_do_not_grow_per_event`, `test_query_count_does_not_grow_with_message_volume`).

Three things worth a look before merging (left as inline comments):
1. **Test coverage gap**: the new Postgres-only raw SQL (`_postgres_candidates`/`_postgres_cursor_clause` in `console/agent_audit/events.py`, `_postgres_latest_visible_outbound_messages` in `api/agent/comms/message_reads.py`) never runs in CI — `config.test_settings` forces SQLite, so only the portable fallback is exercised by `manage.py test`.
2. **Possible regression**: `forAgentId` was dropped from the primary `useAgentRoster` call in `AgentChatPage.tsx`, and there's a code comment describing exactly the bug class this previously fixed; the call site isn't covered by any test since `useAgentRoster` is mocked in `AgentChatPage.test.tsx`.
3. **Perf/correctness nit**: `ChatSidebarGallery`'s new `rows` array isn't memoized the way `ChatSidebar`'s is, so it rebuilds every render and can re-trigger the scroll-to-agent effect unnecessarily.

Everything else (migration using `CREATE INDEX CONCURRENTLY` with `atomic = False`, thumbnail generation/caching service, cache-control `immutable` addition, timeline candidate/hydrate split) looked correct on inspection.

---

PR #1507, opened 2026-08-03
Optimize agent timeline pagination and switch avatar thumbnails to WebP

matt-greathouse:
## Summary

- Optimize timeline candidate selection and hydration, including PostgreSQL-specific pagination queries.
- Cancel stale timeline queries when switching agents.
- Generate 128px WebP avatar thumbnails with transparency support and revisioned cache URLs.
- Expand avatar, timeline, and frontend coverage for the new behavior.

## Testing

- Not run (not requested)

### DIFF
--- a/console/agent_chat/timeline.py
+++ b/console/agent_chat/timeline.py
@@ -115,8 +115,10 @@ class CursorPayload:
             return None
         try:
             value_str, kind, identifier = raw.split(":", 2)
+            if kind == "kanban":
+                kind = "plan"
             return CursorPayload(value=int(value_str), kind=kind, identifier=identifier)
-        except Exception:
+        except (TypeError, ValueError):
             return None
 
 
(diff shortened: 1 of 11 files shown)


console/agent_chat/timeline.py
```diff
@@ -1122,6 +1134,271 @@ def _dt_from_cursor(cursor: CursorPayload) -> datetime:
...
+    if identifier is None:
+        return f" AND {value_sql} {operator} %s", [pivot]
+    return (
+        f" AND ({value_sql} {operator} %s OR "
+        f"({value_sql} = %s AND {identifier_sql} {operator} %s::{identifier_cast}))",
+        [pivot, pivot, identifier],
+    )
+
+
+def _postgres_timeline_candidates(
```
matt-greathouse, line 1110: `_postgres_timeline_candidates` builds a fairly intricate hand-rolled `UNION ALL` (JSONB `->` lookups, `::uuid`/`::varchar` casts, `EXTRACT(EPOCH FROM ...)`) that only runs when `connection.vendor == "postgresql"`. Production uses Postgres (`config/settings.py`), but CI's `DJANGO_SETTINGS_MODULE=config.test_settings` (`.github/workflows/ci.yml:178`) points at `config/test_settings.py`, which overrides `DATABASES` to SQLite. That means this whole code path is never executed against a real Postgres connection by the test suite — `connection.vendor` will be `"sqlite"` in every CI run, so tests always exercise `_portable_timeline_candidates` instead.

The one test that targets this function, `test_postgres_timeline_candidates_use_one_globally_limited_union`, mocks `connection` entirely (`MagicMock`) and only asserts on the generated SQL string/params — it never sends the query to a database, so it can't catch a typo, wrong cast, or type mismatch in the raw SQL. A regression here (this is the query path described as the PR's core "PostgreSQL-specific pagination" optimization) would only surface in production.

Worth adding a Postgres-backed test (or running this specific suite against `config.settings`/a Postgres service in CI) so the raw SQL is actually executed and its results verified before merge, not just its shape.

---

console/agent_chat/timeline.py
```diff
@@ -1659,26 +2020,16 @@ def fetch_timeline_window(
...
-    )
+    if direction == "initial" and not PersistentAgentKanbanEvent.objects.filter(agent=agent).exists():
+        ensure_plan_baseline_event(agent)
 
-    filtered = _filter_by_direction(merged, direction, cursor_payload)
-    truncated = _truncate_for_direction(filtered, direction, limit)
-
-    # Ensure chronological order for presentation
-    truncated.sort(key=lambda env: env.sort_key)
+    candidates = _select_timeline_candidates(agent, cursor_payload, direction, limit)
```
matt-greathouse, line 1838: Now that `fetch_timeline_window` goes through `_select_timeline_candidates`/`_hydrate_timeline_candidates`, the old per-kind queryset builders and merge helpers they replaced — `_messages_queryset` (L975), `_thinking_queryset` (L1032), `_plan_event_queryset` (L1064), `_user_action_event_queryset` (L1092), `_filter_by_direction` (L1578), `_truncate_for_direction` (L1596) — are no longer called from production code (only `_steps_queryset` is still reachable, and only from a unit test). That's ~150 lines of dead code left behind by the refactor; worth deleting them in this PR instead of keeping two parallel, unused implementations around.

---

console/agent_chat/timeline.py
```diff
@@ -1122,6 +1134,271 @@ def _dt_from_cursor(cursor: CursorPayload) -> datetime:
...
+def _coerce_candidate_identifier(identifier: str, id_field: str):
+    if id_field == "seq":
+        return identifier
+    try:
+        return uuid.UUID(identifier)
+    except (TypeError, ValueError):
+        return None
+
+
+def _apply_candidate_cursor(
```
matt-greathouse, line 991: Minor: `CursorPayload.kind` still allows a legacy `"kanban"` value (line 106) for backward compatibility with previously-encoded cursors, and the old `_plan_event_queryset`/`_has_more_before`/`_has_more_after` all special-cased `cursor.kind in {"kanban", "plan"}`. This new tie-break (`kind < cursor.kind` / `kind > cursor.kind`, string-compared against the fixed candidate kind `"plan"`) doesn't account for that alias — a client still holding a stale `kanban`-kind cursor across this deploy would skip the identifier tie-break for plan events at a shared timestamp boundary, risking a duplicated/dropped plan row on that one page. Low impact (only affects an in-flight pagination cursor spanning the deploy), but the old code handled it explicitly and this doesn't.

---

matt-greathouse (commented):
Automated review of the timeline pagination optimization and avatar WebP switch.

The timeline candidate-selection logic (portable ORM path + Postgres raw-SQL path, dual `has_more_*` inference) checks out logically — cursor tie-breaking, truncation direction, and hydration by kind all line up with the added tests. The avatar WebP change (revisioned format string, quality/method constants, transparency handling) also looks correct and is well covered by new tests.

Left 3 comments, the main one being a testing-infrastructure gap: CI runs against SQLite (`config.test_settings`) while production is Postgres, so the new hand-rolled `_postgres_timeline_candidates` raw SQL — the PR's core "PostgreSQL-specific pagination" optimization — is never actually executed against a real database by the test suite. The other two are lower-priority: dead code left over from the refactor, and a minor edge case around a legacy cursor-kind alias.

---

PR #1510, opened 2026-08-04
Fix SQLite source reconciliation loops

matt-greathouse:
## Summary

- reconcile source-derived SQLite mutations by imported JSON path instead of destination-table naming
- allow newly created durable keyed tables while continuing to reject staging tables, literal copies, failed writes, and mutations without bounded reads
- persist stable reconciliation directive hashes in existing SQLite tool-call metadata
- stop repeated focused loops after three total attempts or two successful-but-unverified attempts, then resume normal routing without `sqlite_batch` for one completion
- preserve the multi-array requirement that every keyed source entity must reconcile

## Root cause

Reconciliation inferred source identity from the destination table name. Valid imports into flexibly named durable tables could remain unverified, so the runtime repeatedly focused the agent on `sqlite_batch` across completions.

## Impact

A validated reconciliation immediately restores the full tool set. Repeated unverified or failed attempts are bounded and produce a persisted warning instead of trapping the agent in a SQLite loop. LLM provider routing, retries, failover, and timeout behavior are unchanged.

## Validation

```
uv run python manage.py test --settings=config.test_settings --parallel auto \
  --tag batch_event_processing \
  --tag batch_event_processing_credits \
  --tag batch_sqlite_quality
```

718 tests passed.

### DIFF
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -7,6 +7,7 @@ loop with LLM‑powered reasoning and tool execution using tiered failover.
 """
 from __future__ import annotations
 
+import hashlib
 import json
 import os
 import logging
@@ -1856,13 +1857,149 @@ def _source_reconciliation_parameters(directive: str) -> dict[str, Any]:
             },
             "sql": {
                 "type": "string",
-                "description": directive + " Start with source-derived model DDL/upserts; end with bounded task-relevant rows. No pre-read or copied values.",
+                "description": (
+                    directive
+                    + " Run one focused reconciliation batch: start with source-derived model DDL/upserts and "
+                    "end with bounded task-relevant rows. No pre-read or copied values. After validation, normal "
+                    "tool routing resumes."
+                ),
             },
             "will_continue_work": {"type": "boolean", "const": True},
         },
         "required": ["rows", "sql", "will_continue_work"],
         "additionalProperties": False,
     }
+
+
+SOURCE_RECONCILIATION_MAX_ATTEMPTS = 3
+SOURCE_RECONCILIATION_MAX_UNVERIFIED_SUCCESSES = 2
+SOURCE_RECONCILIATION_METADATA_KEY = "source_reconciliation"
+
+
+@dataclass(frozen=True)
+class _SourceReconciliationAttemptCounts:
+    total: int
+    successful_unverified: int
+
+    @property
+    def exhausted(self) -> bool:
+        return (
+            self.total >= SOURCE_RECONCILIATION_MAX_ATTEMPTS
+            or self.successful_unverified >= SOURCE_RECONCILIATION_MAX_UNVERIFIED_SUCCESSES
+        )
+
+
+def _source_reconciliation_directive_hash(
+    agent: PersistentAgent,
+    directive: str,
+    fresh_tool_call_step_ids: list[str],
+) -> str:
+    inbound = get_current_inbound_message(agent)
+    source_step_ids = []
+    if fresh_tool_call_step_ids:
+        rows = PersistentAgentToolCall.objects.filter(
+            step__agent=agent,
+            step_id__in=fresh_tool_call_step_ids,
+        ).values_list("step_id", "tool_name")
+        source_step_ids = sorted(
+            str(step_id)
+            for step_id, tool_name in rows
+            if is_source_bearing_tool(tool_name)
+        )
+    identity = json.dumps(
+        {
+            "directive": directive,
+            "inbound_message_id": str(inbound.id) if inbound is not None else None,
+            "source_step_ids": source_step_ids,
+        },
+        sort_keys=True,
+        separators=(",", ":"),
+    )
+    return hashlib.sha256(identity.encode("utf-8")).hexdigest()
+
+
+def _source_reconciliation_attempt_counts(
+    agent: PersistentAgent,
+    directive_hash: str,
+) -> _SourceReconciliationAttemptCounts:
+    total = 0
+    successful_unverified = 0
+    rows = PersistentAgentToolCall.objects.filter(
+        step__agent=agent,
+        tool_name="sqlite_batch",
+    ).values_list("status", "result", "display_metadata")
+    for status, result_text, display_metadata in rows:
+        metadata = display_metadata if isinstance(display_metadata, dict) else {}
+        reconciliation = metadata.get(SOURCE_RECONCILIATION_METADATA_KEY)
+        if not isinstance(reconciliation, dict) or reconciliation.get("directive_hash") != directive_hash:
+            continue
+        total += 1
+        if status != PersistentAgentToolCall.Status.COMPLETE:
+            continue
+        try:
+            result = json.loads(result_text)
+        except (TypeError, ValueError):
+            continue
+        if _tool_result_is_success(result):
+            successful_unverified += 1
+    return _SourceReconciliationAttemptCounts(
+        total=total,
+        successful_unverified=successful_unverified,
+    )
+
+
+def _source_reconciliation_release_history(
+    history: list[dict[str, Any]],
+    directive: str,
+) -> list[dict[str, Any]]:
+    released = [dict(message) for message in history]
+    for message in released:
+        content = str(message.get("content") or "")
+        if directive in content:
+            message["content"] = content.replace(directive, "")
+    instruction = (
+        "## SQLite Source Reconciliation Safety Release (CRITICAL)\n\n"
+        "Repeated focused SQLite attempts did not validate this source reconciliation. sqlite_batch is unavailable "
+        "for this completion. Continue with any remaining non-SQL action using the latest bounded result, or clearly "
+        "report that source reconciliation failed. Do not refetch or reconstruct source rows."
+    )
+    system_message = next((message for message in released if message.get("role") == "system"), None)
+    if system_message is None:
+        released.insert(0, {"role": "system", "content": instruction})
+    else:
+        system_message["content"] = str(system_message.get("content") or "") + "\n\n" + instruction
+    return released
+
+
+def _record_source_reconciliation_guard_warning(
+    agent: PersistentAgent,
+    directive_hash: str,
+    counts: _SourceReconciliationAttemptCounts,
+) -> None:
+    _create_agent_system_step_once(
+        agent=agent,
+        description=(
+            "SQLite source reconciliation stopped after repeated focused attempts "
+            f"({counts.total} total, {counts.successful_unverified} successful but unverified)."
+        ),
+        code=PersistentAgentSystemStep.Code.SYSTEM_DIRECTIVE,
+        notes=f"sqlite_source_reconciliation_guard:{directive_hash}",
+    )
+
+
+def _annotate_source_reconciliation_attempt(
+    prepared_calls: list[_PreparedToolExecution],
+    directive_hash: str,
+) -> None:
+    for prepared in prepared_calls:
+        if prepared.tool_name != "sqlite_batch":
+            continue
+        prepared.display_metadata = {
+            **prepared.display_metadata,
+            SOURCE_RECONCILIATION_METADATA_KEY: {
+                "directive_hash": directive_hash,
+            },
+        }
 
 
 CHARTER_PATCH_PARAMETERS = {
@@ -2267,6 +2404,7 @@ class _PreparedToolExecution:
     parallel_ineligible_reason: Optional[str]
     resolved_entry: Optional[ToolCatalogEntry] = None
     pipedream_google_sheets_blocked: bool = False
+    display_metadata: Dict[str, Any] = field(default_factory=dict)
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


api/agent/tools/sqlite_query_quality.py
```diff
@@ -41,6 +41,11 @@
 MANUAL_INSERT_VALUES_RE = re.compile(r'\binsert\s+(?:or\s+\w+\s+)?into\s+"?(?P<name>[a-z_]\w*)"?\s*(?:\(\s*"?[a-z_]\w*"?(?:\s*,\s*"?[a-z_]\w*"?)*\s*\)\s*)?values\b', re.I)
 JSON_FUNCTION_RE = re.compile(r"\bjson_(?:extract|each)\s*\(", re.I)
 JSON_EACH_RE = re.compile(r"\bjson_each\s*\(", re.I)
+SOURCE_JSON_EACH_PATH_RE = re.compile(
+    r"\bjson_each\s*\(\s*(?:(?:[a-z_]\w*)\s*\.\s*)?(?:result_json|analysis_json)\s*,\s*"
+    r"(?P<quote>['\"])(?P<path>\$[^'\"]+)(?P=quote)",
+    re.I,
```
matt-greathouse, line 47: This regex only matches a **bare (optionally alias-qualified) `result_json` / `analysis_json` column as the first `json_each` argument, with a literal path as the second**. Every other way an agent can write the same reconciliation returns no paths at all, and since `prompt_context` now derives `reconciled_entities` from these paths instead of from table names, "no paths" means "never reconciled" — which re-issues the directive forever. That's the loop this PR is trying to close.

Verified against the code on this branch:

```
alias col (supported)    -> ('$.result.posts',)
CTE payload alias        -> ()      # WITH src AS (SELECT result_json AS payload ...) ... json_each(src.payload,'$.result.posts')
subselect first arg      -> ()      # json_each((SELECT result_json FROM __tool_results WHERE ...), '$.result.posts')
root array no path       -> ()      # json_each(t.result_json)     <- payload is a top-level array
json_extract nested      -> ()      # json_each(json_extract(t.result_json,'$.result'))
```

The `root array no path` case is structurally unreachable: the regex requires `,\s*'path'`, so a source whose payload *is* the array can never satisfy the predicate no matter what SQL the agent writes. The other three are formulations the model produces routinely.

Net effect for those shapes: 3 focused attempts, then the new guard strips `sqlite_batch` and the agent reports failure — for a reconciliation that actually succeeded. Suggest at minimum handling the no-path form (treat `json_each(result_json)` as `$`) and resolving one level of CTE/subquery alias back to the payload column, or falling back to the previous table-stem entity when no path can be extracted.

Separately: `source_derived_model_reconciled_paths` re-implements the pending/reconciled walk already in `source_derived_model_reconciled_tables` just below/above it. Two copies of this state machine will drift — worth a single pass returning `dict[table, paths]` that both callers project from.

---

api/agent/tools/sqlite_state.py
```diff
@@ -282,6 +282,43 @@ def get_sqlite_model_table_columns(
...
+        for table_name in table_names:
+            quoted_name = table_name.replace('"', '""')
+            columns = conn.execute(f'PRAGMA table_info("{quoted_name}")').fetchall()
+            if any(int(column[5] or 0) > 0 for column in columns):
+                keyed_tables.add(table_name)
+                continue
+            indexes = conn.execute(f'PRAGMA index_list("{quoted_name}")').fetchall()
+            if any(bool(index[2]) for index in indexes):
+                keyed_tables.add(table_name)
+        return keyed_tables
```
matt-greathouse, line 313: **Case mismatch makes this set unusable for any mixed-case table.** `sqlite_master.name` preserves the case the table was created with, but the set it gets intersected against in `prompt_context` (`source_derived_model_reconciled_tables`) lowercases every name via `_matched_user_table` → `.lower()`. SQLite identifiers are case-insensitive, so these are the same table and the intersection still misses.

Repro on this branch:

```python
conn.execute('CREATE TABLE Reddit_Posts(post_id TEXT PRIMARY KEY, title TEXT)')
get_sqlite_model_tables_with_identity()          # -> {'Reddit_Posts', ...}
source_derived_model_reconciled_tables([sql])    # -> ('reddit_posts',)
# eligible_tables ∩ reconciled_tables == set()
```

`newly_created_keyed_tables` masks this on the batch that creates the table (those names come from `working_table_names`, already lowercased), so the failure only shows up on the common path — table created in an earlier batch, reconciled later. When it does, the source is never marked reconciled and the directive loops until the new guard pulls `sqlite_batch`. Fix is a `.casefold()` here (or at the intersection in `prompt_context.py:5121`).

---

api/agent/core/event_processing.py
```diff
@@ -1865,6 +1871,137 @@ def _source_reconciliation_parameters(directive: str) -> dict[str, Any]:
...
+    )
+
+
+def _source_reconciliation_release_history(
+    history: list[dict[str, Any]],
+    directive: str,
+) -> list[dict[str, Any]]:
+    released = [dict(message) for message in history]
+    for message in released:
+        content = str(message.get("content") or "")
```
matt-greathouse: `content` is not always a string. When `read_file` image attachments are present, `_prepare_multimodal_read_file_completion_request` runs at `event_processing.py:7700` — **before** this release path at 7777 — and `attach_read_file_images_to_messages` rewrites the **last user message** into a list of `{"type": "text"}` / `{"type": "image_url"}` parts. The directive lives in exactly that message (`prompt_context.py:2099` selects it from `user_content`), so the `directive in content` test passes and the list gets replaced by its `str()` repr.

Verified on this branch:

```python
hist = [{'role':'system','content':'sys'},
        {'role':'user','content':[{'type':'text','text':'do it '+d},
                                  {'type':'image_url','image_url':{'url':'data:image/png;base64,AAAA'}}]}]
_source_reconciliation_release_history(hist, d)[1]['content']
# -> "[{'type': 'text', 'text': 'do it '}, {'type': 'image_url', 'image_url': {'url': 'data:image/png;base64,AAAA'}}]"   (a str)
```

The image is destroyed and the base64 data URL is inlined as prompt text. Guard with `isinstance(content, str)` and recurse into the `text` parts when it's a list.

---

api/agent/core/event_processing.py
```diff
@@ -1865,6 +1871,137 @@ def _source_reconciliation_parameters(directive: str) -> dict[str, Any]:
...
+        if directive in content:
+            message["content"] = content.replace(directive, "")
```
matt-greathouse: This assumes `directive` is a single contiguous substring of the message, but `prompt_context.py:2099` builds it by joining **multiple** directives with `"\n"`:

```python
source_reconciliation_directive = "\n".join(
    directive for directive in source_reconciliation_directives if directive in user_content
) or None
```

Those directives are rendered at different points in the prompt tree, so the joined string never appears verbatim and `.replace()` is a no-op:

```python
d2 = 'A-directive\nB-directive'
h2 = [{'role':'user','content':'x A-directive x\ny B-directive y'}]
_source_reconciliation_release_history(h2, d2)[0]['content']
# -> 'x A-directive x\ny B-directive y'   (unchanged)
```

That's the worst failure mode of this feature: `sqlite_batch` is stripped from `request_tools` while the prompt still contains directives ordering the model to run a reconciliation batch. Iterating over the individual directives (or splitting on `"\n"` here) fixes it.

---

api/agent/core/event_processing.py
```diff
@@ -1865,6 +1871,137 @@ def _source_reconciliation_parameters(directive: str) -> dict[str, Any]:
...
+    return hashlib.sha256(identity.encode("utf-8")).hexdigest()
+
+
+def _source_reconciliation_attempt_counts(
+    agent: PersistentAgent,
+    directive_hash: str,
+) -> _SourceReconciliationAttemptCounts:
+    total = 0
+    successful_unverified = 0
+    rows = PersistentAgentToolCall.objects.filter(
```
matt-greathouse, line 1927: This pulls **every** `sqlite_batch` tool call the agent has ever made, including the full `result` column — which the model itself documents as "Raw result or output from the tool call (may be large)" — and then filters in Python, discarding all but the handful that carry this `directive_hash`. It runs once per agent-loop iteration while a directive is active, so a long-lived agent with thousands of `sqlite_batch` rows streams megabytes of result text per iteration to count at most 3 attempts.

`display_metadata` is a `JSONField`, so the filter belongs in the DB:

```python
rows = PersistentAgentToolCall.objects.filter(
    step__agent=agent,
    tool_name="sqlite_batch",
    display_metadata__source_reconciliation__directive_hash=directive_hash,
).values_list("status", "result")
```

An index-friendly bound on `step__created_at` (or ordering + slicing to `SOURCE_RECONCILIATION_MAX_ATTEMPTS + 1`) would help too — the counters only ever need to distinguish "< limit" from "≥ limit".

---

api/agent/tools/sqlite_state.py
```diff
@@ -282,6 +282,43 @@ def get_sqlite_model_table_columns(
...
+            ).fetchall()
+        }
+        keyed_tables = set()
+        for table_name in table_names:
+            quoted_name = table_name.replace('"', '""')
+            columns = conn.execute(f'PRAGMA table_info("{quoted_name}")').fetchall()
+            if any(int(column[5] or 0) > 0 for column in columns):
+                keyed_tables.add(table_name)
+                continue
+            indexes = conn.execute(f'PRAGMA index_list("{quoted_name}")').fetchall()
```
matt-greathouse, line 310: Two smaller things in this identity probe:

1. `index_list` column 2 is `unique`, but column 4 is `partial`. A partial unique index (`CREATE UNIQUE INDEX ix ON t(a) WHERE a IS NOT NULL`) doesn't give the table row identity, yet it's accepted here — I see `part_idx` come back as keyed in a local repro. Worth adding `and not index[4]`.

2. The table filter (`name NOT LIKE 'sqlite_%' AND substr(name,1,2) != '__'`) is a third, independent definition of "model table", diverging from `_is_named_model_table`'s `NON_MODEL_TABLE_PREFIXES` / `NON_MODEL_TABLE_SUFFIXES` (`temp_`, `staging_`, `_tmp`, …). Today that's inert — the intersection in `prompt_context` is against `source_derived_model_reconciled_tables`, which already applies `_is_named_model_table` — but the docstring says "durable model tables", and the extra names cost two PRAGMA round-trips each on every prompt render for tables that can never match.

On cost generally: this opens a connection and issues 2 PRAGMAs per table on **every** `_render_prompt_context_once`, even when there are no source reconciliations to evaluate. Deferring it behind `if source_reconciliations:` would keep it off the hot path.

---

api/agent/core/event_processing.py
```diff
@@ -1865,6 +1871,137 @@ def _source_reconciliation_parameters(directive: str) -> dict[str, Any]:
     }
 
 
+SOURCE_RECONCILIATION_MAX_ATTEMPTS = 3
```
matt-greathouse, line 1874: These budgets count *attempts*, not *progress*, and they persist across processing runs (the new `test_failed_source_reconciliation_guard_persists_across_followup_run` confirms that's intended). Since the directive is issued with `fixed_continue=True` and the schema pins `will_continue_work: True`, an agent legitimately reconciling a large source across several bounded batches — page 1, page 2, page 3 — trips `successful_unverified >= 2` on the second batch and loses `sqlite_batch` for the rest of the completion, even though each batch succeeded.

The hash only resets on a new inbound message or a new source-bearing tool call, neither of which happens mid-reconciliation. Consider keying "unverified success" on the batch making no forward change (e.g. identical normalized SQL, or no new rows in the target table) rather than on the raw count, so multi-batch reconciliation isn't indistinguishable from a spin.

---

matt-greathouse (commented):
Reviewed the 7 changed files. The attempt-budget guard is a sound safety net and the release path is well-structured — the escape hatch, the once-only system step, the `DatabaseError` handling, and the cross-run persistence all look right. `tests/unit/test_sqlite_batch.py::SqliteBatchQualityTests` passes locally (56 tests).

My concern is with the *other half* of the change: the reconciliation predicate got materially narrower at the same time the guard was added, so the guard will fire on reconciliations that actually worked.

**Likely to cause loops rather than fix them**
1. `SOURCE_JSON_EACH_PATH_RE` recognizes only a bare `alias.result_json` first argument with a literal path. CTE aliases, subselects, `json_each(json_extract(...))`, and top-level-array payloads (`json_each(result_json)` with no path) all yield `()` — and the last one is structurally unreachable, so those sources can *never* be marked reconciled. Verified with a repro in the inline comment.
2. `get_sqlite_model_tables_with_identity()` returns `sqlite_master` names verbatim while the set it's intersected with is lowercased, so any mixed-case table fails the new eligibility gate on every batch after the one that created it.

Both land in the same place: 3 focused attempts, then `sqlite_batch` is stripped and the agent reports failure on a successful reconciliation.

**Correctness in the release path**
3. `_source_reconciliation_release_history` stringifies list-valued `content`, which destroys `read_file` image blocks (multimodal attachment runs earlier in the loop and targets the same user message the directive lives in).
4. It also can't strip a multi-directive string — `prompt_context` joins directives with `"\n"`, which never appears verbatim — leaving the prompt demanding SQL while the tool is gone.

**Also flagged:** an unbounded scan of every historical `sqlite_batch` `result` per loop iteration where a JSONField filter would do; partial unique indexes counted as identity; and the attempt budget counting attempts rather than progress, which penalizes legitimate multi-batch reconciliation.

**Test gaps:** `get_sqlite_model_tables_with_identity` has no direct coverage (unique-index, partial-index, and casing paths are all untested), and `_source_reconciliation_release_history` isn't tested with list content, a missing system message, or a multi-directive string. Adding a case for each of the SQL formulations in comment 1 would pin down the intended breadth of the predicate.

---

PR #1519, opened 2026-08-08
Show human-readable names in Discord agent interactions

matt-greathouse:
## Summary

- Improve Discord message and reaction handling to preserve and display human-readable user names.
- Update Discord channel subscriptions, outbound content policy, prompt context, and native system skill defaults.
- Expand Discord and responsibility-boundary eval scenarios and regression coverage.

## Testing

- Added and updated unit, integration, and eval tests covering Discord bot behavior, event processing, link references, outbound content policy, and trigger subscriptions.
- Not run (not requested).


api/agent/core/prompt_context.py
```diff
@@ -2758,6 +2764,15 @@ def _build_contacts_block(
                 endpoint_address = endpoint.address
         if not endpoint_address:
             continue
+        if endpoint_channel == CommsChannel.DISCORD:
+            raw_payload = msg.raw_payload if isinstance(msg.raw_payload, dict) else {}
+            guild_id = str(raw_payload.get("discord_guild_id") or "").strip()
+            guild_name = str(raw_payload.get("discord_guild_name") or "").strip()
+            channel_name = str(raw_payload.get("discord_channel_name") or "").strip().lstrip("#")
+            if not guild_id or not channel_name:
```
matt-greathouse: This `continue` silently drops a Discord conversation from the recent-contacts block whenever the message's `raw_payload` lacks `discord_guild_id`/`discord_channel_name`, with no fallback to the raw endpoint address (unlike every other channel type above).

Concretely: `_prepare_inactive_discord_auto_reply` (api/services/discord_bot.py) creates the outbound message with `raw_payload={"kind": INACTIVE_AUTO_REPLY_KIND, **metadata}` — no guild fields — and status `QUEUED`. If the subsequent `send_channel_message()` webhook call in `deliver()` then fails, the `except` block only updates `latest_status`/`latest_error_message`, never `raw_payload`, so the row is permanently stuck without `discord_guild_id`. From then on this loop computes `guild_id=''` and skips the conversation entirely — a regression from the previous behavior where the raw endpoint address was always shown.

Worth either (a) falling back to the endpoint address when guild/channel fields are missing, like the other channels do, or (b) backfilling guild fields onto the message when the deliver() failure handler runs.

---

api/agent/core/prompt_context.py
```diff
@@ -4642,6 +4657,104 @@ def _format_discord_reply_context(raw_payload: Mapping[str, Any]) -> str:
...
+            if key not in _DISCORD_CHANNEL_ID_KEYS
+            and key not in {"agent_id", "subscription_id"}
+            and not (
+                key == "id"
+                and "channel_name" in value
+                and ("guild_id" in value or "guild_name" in value)
+            )
+        }
+        if channel_id:
+            for key, item in channel_names.get(channel_id, {}).items():
```
matt-greathouse, line 4725: `_sanitize_discord_tool_payload_for_prompt` always strips `channel_id`/`discord_channel_id` from historical Discord tool calls/results (line ~4718), then only restores `guild_id`/`guild_name`/`channel_name` if that channel id still has a live entry in `_discord_subscription_names_by_channel_id(agent)`. If the channel was later unsubscribed (or the subscription id changed), `channel_names.get(channel_id, {})` returns `{}` and the sanitized payload ends up with no channel/guild identification at all — strictly worse than the pre-PR behavior, which still showed the raw `channel_id`.

Consider keeping the original `channel_id` as a last-resort fallback when no current subscription matches, rather than dropping it unconditionally.

---

api/agent/tools/discord_channel_subscriptions.py
```diff
@@ -115,33 +136,49 @@ def execute_discord_channel_subscriptions(agent: PersistentAgent, params: Dict[s
...
 
         if action == "disable":
-            subscription_id = str(params.get("subscription_id") or "").strip()
-            if not subscription_id:
-                return {"status": "error", "message": "subscription_id is required for disable."}
+            guild_id = str(params.get("guild_id") or "").strip()
+            channel_name = str(params.get("channel_name") or "").strip()
+            if not guild_id or not channel_name:
+                return {"status": "error", "message": "guild_id and channel_name are required for disable."}
+            subscription = resolve_active_subscription(
```
matt-greathouse: The `disable` action lost its idempotency here. Previously it took a `subscription_id` and called `disable_subscription()` directly, which just sets `status=DISABLED` regardless of current status — so disabling twice was harmless. Now it first calls `resolve_active_subscription()`, which filters on `status=ACTIVE` (api/services/discord_bot.py:795) and raises `DiscordBotIntegrationError("No active native Discord subscription matched that channel.")` if nothing active matches.

So a retry of the same disable call (e.g. after a timeout, or the agent/user repeating "stop that channel") now surfaces as an error instead of succeeding as a no-op. Worth confirming this is intentional, since an LLM-driven tool retrying on an ambiguous prior result is a fairly likely scenario here.

---

api/services/discord_messages.py
```diff
@@ -70,7 +70,7 @@ def ensure_conversation_participant(
 
 
 def display_name_for_channel(channel_id: str, channel_name: str = "") -> str:
```
matt-greathouse: Minor: `display_name_for_channel` (and `discord_channel_source_label`, which just delegates to it) now ignore the `channel_id` parameter entirely — the old `f"Discord {channel_id}"` fallback became a fixed `"Discord channel"` string. Not a bug, but `channel_id` is now a dead parameter threaded through every call site in this file and `discord_bot.py`; could drop it from both signatures for clarity, or note why it's kept.

---

matt-greathouse (commented):
Automated code review (`/code-review`, medium effort) on `discord-human-names`. All findings below were manually traced against the diff before posting.

**4 issues found**, all regressions introduced by this PR's refactor away from raw Discord IDs:
1. `_build_contacts_block` can permanently drop a Discord conversation from recent contacts if an outbound message's `raw_payload` never gets guild fields (concretely: a failed inactive-auto-reply delivery) — no fallback to the raw endpoint address like other channels have.
2. `_sanitize_discord_tool_payload_for_prompt` can strip all channel/guild identification from historical Discord tool calls if the channel was later unsubscribed, with no fallback to the raw id.
3. The `discord_channel_subscriptions` tool's `disable` action lost idempotency — retrying a disable on an already-disabled channel now errors instead of no-op succeeding.
4. Minor: `channel_id` became a dead parameter in `display_name_for_channel`/`discord_channel_source_label`.

Line comments have the specifics and suggested fixes on each.

---

api/agent/tools/send_discord_message.py
```diff
@@ -95,37 +92,30 @@ def execute_send_discord_message(agent: PersistentAgent, params: Dict[str, Any])
...
-                channel_name=channel_name,
-                guild_id=guild_id,
-            )
-        except DiscordBotIntegrationError as exc:
-            return {"status": "error", "message": str(exc)}
-        channel_id = subscription.channel_id
-    elif not channel_id:
-        return {"status": "error", "message": "channel_id or channel_name is required."}
+    try:
+        subscription = resolve_active_subscription(
```
matt-greathouse, line 96: **A renamed Discord channel now breaks every agent reply, with no fallback.**

The two sides of this handshake read different sources of truth for the channel name:

- The prompt feeds the model the **live** name from the inbound gateway payload — `prompt_context.py:5707` emits `discord_channel_name` from `raw_payload`, which `_ingest_gateway_message_for_subscription` populates from `message.channel_name` on each event (`discord_bot.py:985`).
- `resolve_active_subscription` matches against the **stored** `subscription.channel_name` (`discord_bot.py:807-811`).

Nothing re-syncs the stored name on the inbound path. `record_message()` only writes `last_message_at` / `last_error` (`api/models.py:9973`), and `_validate_subscription_channel` fetches the channel by ID without persisting the fresh name. Only `ensure_subscription` updates it, and that runs solely on explicit (re)subscribe.

So after a user renames `#updates` → `#announcements`: inbound still works (subscription lookup is by `channel_id`), the model is told the channel is `#announcements`, and every `send_discord_message` / `add_discord_reaction` fails with *"No active native Discord subscription matched that channel."* The `channel_id` escape hatch that previously covered this was removed from all three tools in this PR, so the agent has nothing to fall back to — and re-running `ensure` requires the agent to first notice the mismatch.

Cheapest fix: refresh the stored name during ingestion, e.g. in `_finalize_gateway_subscription_delivery` where `record_message()` is already called —

```python
if message.channel_name and subscription.channel_name != message.channel_name:
    subscription.channel_name = message.channel_name
    subscription.save(update_fields=["channel_name", "updated_at"])
```

A test covering rename-then-reply would lock this down; the current suite only exercises names that already agree.

---

api/agent/tools/discord_channel_subscriptions.py
matt-greathouse: **Duplicate channel names become permanently unreachable — including via `disable`.**

Discord allows two text channels with the same name in different categories of one guild, and the uniqueness constraint is `(agent, guild, channel_id)` (`api/models.py:9960`), so both can hold active subscriptions. With `channel_id` and `subscription_id` removed from every parameter list, `resolve_active_subscription` has no disambiguator left and raises for `send`, `disable`, and `ensure` alike.

The error text makes the trap explicit:

> "That channel name matches more than one active subscription in the selected server. Rename or remove the duplicate channels before trying again."

"Remove" means `disable` — which this branch routes through the same `resolve_active_subscription` and which therefore fails identically. The agent cannot dig itself out; it needs a human in the console.

Note `_resolve_text_channel_in_guild` rejects ambiguity at `ensure` time, so this state can't be created through the agent — but it's still reachable two ways: `console/discord_api.py` still passes `channel_id` to `ensure_subscription`, and a rename can collide two previously-distinct subscriptions after the fact.

`test_send_message_tool_rejects_ambiguous_channel_name` (tests/unit/test_discord_bot.py:2002) encodes this dead end as expected behavior. Worth reconsidering: keeping an optional `channel_id` on `disable` only (agent-facing surface stays name-first, but the escape hatch survives) would resolve it without reintroducing IDs into the common path.

---

api/agent/core/prompt_context.py
matt-greathouse: **This `continue` drops the contact entirely, and there is no longer a fallback path that re-adds it.**

The same PR excludes `CommsChannel.DISCORD` from both `agent_eps_qs` (line 2640) and `user_eps_qs` (line 2676), so this loop is now the *only* way a Discord channel reaches the contacts block. When `raw_payload` lacks `discord_guild_id` or `discord_channel_name`, the channel vanishes rather than degrading to a partial label.

Concrete case: `_prepare_inactive_discord_auto_reply` (`discord_bot.py:1396`) persists its message with `raw_payload={"kind": INACTIVE_AUTO_REPLY_KIND, **metadata}` — no guild fields. Successful delivery replaces the payload wholesale with the full one (`discord_bot.py:1367`), so this only bites while the message is `QUEUED` or `FAILED`. That's narrow, but it's exactly the window where the agent most needs to see the channel it failed to reach.

Since `msg.conversation.display_name` is already maintained as `#channel` by `_finalize_gateway_subscription_delivery`, a degraded label costs little:

```python
if not guild_id or not channel_name:
    label = getattr(msg.conversation, "display_name", "") or channel_name
    if not label:
        continue
    endpoint_address = label
else:
    ...
```

Also worth confirming: any other producer of Discord `raw_payload` that omits `discord_guild_id` hits the same silent drop.

---

api/agent/core/prompt_context.py
```diff
@@ -1181,12 +1181,16 @@ def _get_shared_channel_names(agent: PersistentAgent) -> dict:
...
+        channel_id: channel_name
+        for channel_id, channel_name in subscriptions.filter(agent=agent).values_list("channel_id", "channel_name")
+        if str(channel_name or "").strip()
+    }
     shared: dict = {}
     for other_id, channel_id in (
         subscriptions.filter(channel_id__in=list(own)).exclude(agent_id=agent.id).values_list("agent_id", "channel_id")
     ):
-        shared.setdefault(other_id, []).append(f"#{str(own.get(channel_id) or channel_id).lstrip('#')}")
+        shared.setdefault(other_id, []).append(f"#{str(own[channel_id]).lstrip('#')}")
```
matt-greathouse, line 1193: **Filtering blank names out of `own` drops the whole channel from the shared-peer map, not just its label.**

`own` is now built with `if str(channel_name or "").strip()`, and the query below filters on `channel_id__in=list(own)` — so a subscription with an empty `channel_name` is excluded from the *membership* check as well. A peer co-subscribed to that channel is no longer reported as already seeing the message, which is precisely the duplicate-notification case this helper's docstring says it exists to prevent. The previous `own.get(channel_id) or channel_id` fallback degraded the label but kept the peer flagged.

`channel_name` is `blank=True` on the model (`api/models.py:9951`), so empty values are representable — legacy rows predating name capture, or a subscription created before the canonical name resolved.

Keeping the full `own` set for the membership query and skipping only the label is a small change:

```python
own = dict(subscriptions.filter(agent=agent).values_list("channel_id", "channel_name"))
for other_id, channel_id in (...):
    name = str(own.get(channel_id) or "").strip()
    if not name:
        continue  # or fall back to a generic "#unknown-channel"
    shared.setdefault(other_id, []).append(f"#{name.lstrip('#')}")
```

That keeps IDs out of the prompt (the goal here) while preserving the dedup signal. Whether a nameless channel should surface *some* placeholder is a judgment call — but silently unflagging the peer looks unintended.

---

api/services/discord_messages.py
```diff
@@ -70,7 +70,7 @@ def ensure_conversation_participant(
...
-    return f"#{channel_name.lstrip('#')}" if channel_name else f"Discord {channel_id}"
+    return f"#{channel_name.lstrip('#')}" if channel_name else "Discord channel"
```
matt-greathouse, line 73: **Constant fallback makes distinct unnamed channels indistinguishable, and `channel_id` is now unused.**

`display_name_for_channel` feeds `PersistentAgentConversation.display_name` (`get_or_create_discord_conversation`) and, via `discord_channel_source_label`, the `source_label` on stored messages. With the fallback collapsed to the literal `"Discord channel"`, two unnamed subscriptions in different guilds render identically in the console conversation list — the user sees two rows labeled "Discord channel" with nothing to tell them apart. The old `f"Discord {channel_id}"` was ugly but distinguishing.

Worth noting the surface this affects: these are console/user-facing labels, not the agent prompt (which the tool sanitizers now cover separately), so an ID here doesn't leak into the model context the way it would in a tool result.

`channel_id` is also now an unused parameter on both this function and `discord_channel_source_label`. If the constant fallback is intentional, drop the parameter and update the ~15 call sites; if not, a suffixed fallback like `f"Discord channel ({channel_id[-4:]})"` keeps rows distinguishable without printing the full snowflake.

---

matt-greathouse (commented):
## Code review

Reviewed all 16 files against base `dc8d4af`. The direction is good — swapping raw snowflakes for `guild_id` + human-readable `channel_name` across tool schemas, tool results, prompt history, and the contacts block, with the `_sanitize_discord_tool_*_for_prompt` helpers scrubbing IDs out of replayed history. The sanitizers memoize their subscription lookup correctly and fall through safely on non-JSON results.

The main risk is that removing `channel_id` from **all three** tools left no fallback for the cases where name-based resolution can't succeed. Five findings, highest severity first:

1. **Renamed channel breaks every agent reply** (`send_discord_message.py`) — the prompt shows the *live* channel name from the gateway payload, but resolution matches the *stored* name, and nothing re-syncs it on the inbound path. Post-rename, all sends and reactions fail. This is the one I'd fix before merge.
2. **Duplicate channel names are permanently unreachable** (`discord_channel_subscriptions.py`) — `disable` now resolves by name too, so the error's own advice ("remove the duplicate") is impossible to follow from the agent side.
3. **Contacts entries silently dropped** when `raw_payload` lacks guild fields (`prompt_context.py:2772`) — Discord was simultaneously excluded from both endpoint querysets, so nothing re-adds them.
4. **Blank stored `channel_name` drops the channel from the shared-peer map** (`prompt_context.py:1193`), losing the duplicate-notification guard rather than just the label.
5. **`display_name_for_channel` fallback collapsed to a constant** (`discord_messages.py:73`) — unnamed channels are now indistinguishable in the console; `channel_id` is an unused parameter.

Things I checked that are fine: `send_channel_message` already required an active subscription, so requiring one in the tool isn't a new constraint; the new `resolve_active_subscription` calls sit inside the existing exception handlers; `own[channel_id]` can't `KeyError` since the filter derives from the same dict; `ensure_subscription` always returns a `subscription` key; and `console/discord_api.py` still passes `channel_id`, which the service signature still accepts.

One caveat on method: the Django test suite wouldn't run in my environment (the runner hung during DB setup), so these are from reading the code rather than observed failures. `py_compile` passes on all changed modules. Worth running the suite locally before merge.

---

PR #1520, opened 2026-08-08
Add Discord user verification and identity linking

matt-greathouse:
## Summary

- Add Discord identity storage, OAuth callback handling, and profile management.
- Verify Discord users during event processing and expose identity context to agents.
- Add native Discord setup UI and evaluation coverage.

## Testing

- Added and updated focused unit tests for profile APIs, Discord bot behavior, event processing, and native Discord scenarios.
- Not run (not requested)

### DIFF
--- a/frontend/src/hooks/useWindowMessage.ts
+++ b/frontend/src/hooks/useWindowMessage.ts
@@ -2,12 +2,32 @@ import { useEffect } from 'react'
 
 export function useWindowMessage<T extends { type?: unknown }>(type: string, onMessage: (message: T) => void) {
   useEffect(() => {
+    const deliver = (message: T) => {
+      if (message?.type === type) {
+        onMessage(message)
+      }
+    }
     const handleMessage = (event: MessageEvent<T>) => {
-      if (event.origin === window.location.origin && event.data?.type === type) {
-        onMessage(event.data)
+      if (event.origin === window.location.origin) {
+        deliver(event.data)
+      }
+    }
+    const handleStorage = (event: StorageEvent) => {
+      if (!event.key?.startsWith('gobii:oauth_complete:') || !event.newValue) {
+        return
+      }
+      try {
+        deliver(JSON.parse(event.newValue) as T)
+        localStorage.removeItem(event.key)
+      } catch {
+        // Ignore unrelated or malformed completion entries.
       }
     }
     window.addEventListener('message', handleMessage)
-    return () => window.removeEventListener('message', handleMessage)
+    window.addEventListener('storage', handleStorage)
+    return () => {
+      window.removeEventListener('message', handleMessage)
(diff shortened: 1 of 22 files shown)


frontend/src/hooks/useWindowMessage.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function useWindowMessage<T extends { type?: unknown }>(type: string, onMessage: (message: T) => void) {
+  useEffect(() => {
+    const deliver = (message: T) => {
+      if (message?.type === type) {
+        onMessage(message)
+      }
+    }
+    const handleMessage = (event: MessageEvent<T>) => {
+      if (event.origin === window.location.origin) {
```
matt-greathouse: **Bug:** this same-origin check defeats the cross-origin support this commit is meant to add.

`event.origin` here is the origin of the *sender* — the Discord OAuth callback page, served from the fixed `DISCORD_OAUTH_REDIRECT_URI` host. `window.location.origin` is the *opener's* origin (captured server-side as `return_origin`/`target_origin` in `console/discord_api.py`). In exactly the cross-origin scenario this PR's commit title calls out (start and callback on different hosts), these will never be equal, so the completion message is silently dropped and the UI never reflects the successful link until a manual reload.

The `storage` fallback (below) has the same problem for a different reason: `localStorage.setItem` in the callback page writes to *its own* origin's storage, and the `storage` event only fires in other same-origin windows — never cross-origin — so it can't rescue this case either.

Since the server already signs and passes back the legitimate opener origin (`target_origin`), consider validating `event.origin` against that value (e.g. threaded through as a query param or embedded in the payload) instead of `window.location.origin`.

---

frontend/src/hooks/useWindowMessage.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect } from 'react'
```
matt-greathouse, line 1: **Duplication:** `frontend/src/components/mcp/NativeIntegrationShared.tsx` already implements this exact pattern — a `message` listener gated on `event.origin === window.location.origin` plus a `storage` listener keyed by a localStorage prefix — for the MCP OAuth flow. This new hook re-implements it rather than extracting/reusing the existing logic, so both copies now carry the same latent same-origin bug independently, and any future fix has to be applied twice.

---

matt-greathouse (commented):
Automated code review via `/code-review`.

Traced the Discord identity OAuth flow end-to-end (state signing/validation bound to `user.pk`, `IntegrityError` handling on colliding `discord_user_id`, config-authority resolver changes, event-processing feedback-burst filtering, and the new `UserDiscordIdentity` model/migration) — all sound. One real bug found in the new cross-origin `postMessage` completion handling, plus a duplication note. Details inline.

---

PR #1521, opened 2026-08-08
Add Discord embed support for agent messages

matt-greathouse:
## Summary

- Add Discord embed generation and delivery for outbound agent messages
- Preserve and render link references across prompt context, timelines, and agent chat
- Add a Discord embed card to the frontend message timeline
- Extend Discord bot behavior, eval scenarios, and outbound content policy coverage

## Testing

- Added and updated focused unit, integration, eval, and frontend component tests
- Not run (not requested)


api/agent/tools/send_discord_message.py
```diff
@@ -86,15 +87,19 @@ def execute_send_discord_message(agent: PersistentAgent, params: Dict[str, Any])
     guild_id = str(params.get("guild_id") or "").strip()
     body = str(params.get("message") or "").strip()
     attachment_paths = params.get("attachments")
+    try:
+        embeds = normalize_discord_embeds(resolve_nested_link_references(params.get("embeds"), agent))
```
matt-greathouse: `LinkReferenceResolutionError` (raised inside `resolve_nested_link_references`) is a `ValueError` subclass, so this local `except ValueError` catches it before the outer `@handle_link_reference_errors` decorator can run. That means a bad `$[link:...]` token inside an embed field returns `{"status": "error", ...}` **without** `retryable: True`, unlike the identical error raised from the `message` field a few lines below (which does propagate through the decorator). Downstream, `_normalize_error_result`/`_filter_tools_after_non_retryable_result` will treat this as non-retryable and strip `send_discord_message` from the available tools for the rest of the batch, even though the agent could just retry with a corrected token.

Consider narrowing the catch here, e.g. catch `LinkReferenceResolutionError` separately (returning `link_reference_error_response(exc)`) before the generic `ValueError` catch for `normalize_discord_embeds`'s own validation errors.

---

api/agent/tools/send_discord_message.py
```diff
@@ -86,15 +87,19 @@ def execute_send_discord_message(agent: PersistentAgent, params: Dict[str, Any])
     guild_id = str(params.get("guild_id") or "").strip()
     body = str(params.get("message") or "").strip()
     attachment_paths = params.get("attachments")
+    try:
```
matt-greathouse, line 91: Embeds only go through `resolve_nested_link_references` (link-reference resolution), never `substitute_variables_with_filespace` like `body` does two lines below. Every other channel sender (email/sms/mcp/peer_dm/web_chat) applies `substitute_variables_with_filespace` to its outgoing content, which resolves both link references and `$[VAR]` agent-variable/filespace substitution. Here, an agent putting a `$[VAR]` placeholder or a filespace image reference in an embed title/description/field will get the literal placeholder text sent to Discord instead of the resolved value/image.

Worth applying `substitute_variables_with_filespace` (or equivalent variable substitution) to embed text fields for parity with the other send tools.

---

api/agent/core/prompt_context.py
```diff
@@ -5615,6 +5619,11 @@ def _format_web_party(address: str, endpoint_id: UUID | None) -> str:
                 discord_message_id = str(raw_payload.get("discord_message_id") or "").strip()
                 if discord_message_id:
                     components["discord_message_id"] = discord_message_id
+                embed_context = format_discord_embeds(
+                    raw_payload.get("discord_sent_embeds") if m.is_outbound else raw_payload.get("discord_embeds")
+                )
+                if embed_context:
+                    components["discord_embeds"] = embed_context
```
matt-greathouse, line 5626: The new `discord_embeds` component name isn't in `COMPONENT_WEIGHTS` (defined further down in this function), so it silently falls back to the default weight of 1 via `COMPONENT_WEIGHTS.get(component_name, 1)`. Its siblings added in this same diff — `discord_channel_id`/`discord_message_id` (weight 3) and `discord_reply_context` (weight 2) — were all deliberately given higher weights. As-is, embed content (often the only structured info in a message, e.g. status fields) gets pruned/shrunk as aggressively as low-priority scratch content when the prompt builder needs to fit budget.

Consider adding an entry like `"discord_embeds": 2` to `COMPONENT_WEIGHTS`.

---

frontend/src/components/agentChat/MessageEventCard.test.tsx
```diff
@@ -92,11 +92,81 @@ describe('MessageEventCard reply context', () => {
     expect(screen.getByText('honestly maybe like a 30.')).toBeInTheDocument()
   })
 
+  it('shows embed, attachment, and unavailable reply fallbacks supplied by the timeline', () => {
```
matt-greathouse, line 95: AGENTS.md says "Do not add front-end tests unless specifically requested." This PR adds two new front-end test cases here (embed/attachment/unavailable reply fallbacks, and top-level embeds without text). Flagging in case these weren't specifically requested — if they were, feel free to disregard.

---

matt-greathouse (commented):
Reviewed the Discord embed support changes (branch `discord-embed-support`). Found 4 issues, all verified against the code: an error-handling bug where a `ValueError` catch inadvertently swallows `LinkReferenceResolutionError` and drops its `retryable` marker, missing `$[VAR]`/filespace variable substitution in embed text (inconsistent with every other send-tool), a new `discord_embeds` prompt-history component missing from `COMPONENT_WEIGHTS` (so it gets pruned as aggressively as low-priority content unlike its sibling fields added in the same diff), and two new front-end tests that may conflict with the repo's "no front-end tests unless requested" rule. Details inline.

---

PR #1525, opened 2026-08-09
Improve asynchronous history and step compaction

matt-greathouse:
## Summary

- Add asynchronous history compaction task handling and compaction-specific exceptions
- Refine prompt context, event processing, link references, and step compaction behavior
- Update token usage, agent chat, Discord, multimodal, system skill, and compaction test coverage
- Add effort calibration coverage and configuration support

## Testing

- Not run (not requested)


api/agent/core/history_compaction.py
```diff
@@ -0,0 +1,168 @@
...
+            agent.id,
+            exc_info=True,
+        )
+        lease_token = ""
+
+    routing_profile_id = str(routing_profile.id) if routing_profile is not None else None
+    try:
+        from ..tasks.history_compaction import compact_agent_history_task
+
+        compact_agent_history_task.apply_async(
```
matt-greathouse, line 157: `enqueue_history_compaction` is now called unconditionally from `build_prompt_context` (prompt_context.py:2319) with no surrounding try/except at the call site. Here, `apply_async` is only guarded by `except (CeleryError, KombuOperationalError)`. Any other exception raised while acquiring the lease or publishing to the broker (e.g. a raw connection error not wrapped as one of those types) will propagate out of `build_prompt_context`, into `_run_agent_loop`'s try/except (event_processing.py:7526-7576), which logs it and then re-raises — aborting the entire agent turn.

Previously, compaction failures were fully isolated (`ensure_comms_compacted`/`ensure_steps_compacted` caught summarizer exceptions internally), so this is a real behavior change: a routine best-effort scheduling hiccup can now fail an otherwise-healthy prompt build. Consider wrapping the whole `enqueue_history_compaction` call (or at least the `apply_async`) in a broader `except Exception` so scheduling failures stay best-effort as intended.

---

api/agent/tasks/history_compaction.py
```diff
@@ -0,0 +1,103 @@
...
+                    f"Routing profile {routing_profile_id} is unavailable"
+                )
+
+        summarizer_kwargs = {
+            "agent": agent,
+            "routing_profile": routing_profile,
+            "eval_run_id": eval_run_id,
+        }
+        safety_identifier = agent.user_id
+        ensure_steps_compacted(
```
matt-greathouse, line 77: The coalescing lease is only refreshed at task start (`history_compaction.py:42`) and again on retry (`:90`), but not while `ensure_steps_compacted`/`ensure_comms_compacted` are doing the actual LLM summarization work (Phase 2, which can be slow). With the default `PA_HISTORY_COMPACTION_LEASE_SECONDS = 3600`, an agent with a large backlog or a slow provider could have this task run past the lease TTL without hitting a `CompactionSummaryError`/`DatabaseError` (the only paths that refresh the lease). The Redis lease key then expires mid-run, and the next `build_prompt_context()` call for the same agent will acquire a fresh lease and dispatch a second `compact_agent_history_task`, redundantly re-running (and re-paying for) the same summarization before losing the Phase-3 optimistic-concurrency race.

Not a correctness/data-corruption issue (Phase 3's id-diff check still protects against double-writing a snapshot), but it reintroduces duplicate LLM work that the lease was meant to coalesce away. Worth periodically refreshing the lease during the Phase 2 work itself, not just around retries.

---

api/agent/core/compaction.py
```diff
@@ -317,11 +306,6 @@ def llm_summarise_comms(
...
         return response.choices[0].message.content.strip()
-    except Exception:
-        # Log and fall back to deterministic fallback so callers are not
-        # blocked by transient LLM/network issues.
-        logger.exception("LiteLLM summarisation failed – falling back to fallback summariser")
-        return _default_summarise(previous, messages)
-
-# Re-export for convenience – avoids changing existing imports elsewhere
-from .step_compaction import ensure_steps_compacted  # noqa: E402, isort:skip 
+    except Exception as exc:
```
matt-greathouse: Previously, an exception here fell back to `_default_summarise` locally (no retry, no task failure). Now it's re-raised as `CompactionSummaryError`, which `compact_agent_history_task` treats as retryable (up to 2 retries at 5s/20s). Since this still catches bare `Exception`, a non-transient bug (e.g. a `TypeError`/`AttributeError` in payload formatting before the network call even happens) gets the same retry treatment as a genuine transient LLM/network failure — burning ~25s and two more lock-acquisition windows before finally giving up, and masking the real defect as an infra blip in the "exhausted retries" log line.

Consider narrowing this to catch expected transient failure types (or re-raising unexpected exception types directly) so retries are reserved for things retrying can actually fix.

---

api/agent/core/compaction.py
```diff
@@ -131,6 +125,18 @@ def ensure_comms_compacted(
             span.set_attribute("compaction.skipped", True)
             return
 
+        current_message_ids = frozenset(
```
matt-greathouse, line 129: Minor: this Phase-3 optimistic-concurrency check (capture ids before the LLM call, re-fetch and diff after, bail out on mismatch) is duplicated almost verbatim in `step_compaction.py` (`captured_step_ids`/`current_step_ids` around line 215-265). Since both streams now share the same async task/lease infra, a small shared helper for "diff captured ids vs. current ids under lock" would keep the race-detection logic from drifting between the two copies if it's ever changed in one place but not the other.

---

matt-greathouse (commented):
Automated review of the async history-compaction refactor. Overall the coalescing-lease + async-task design is sound and the Phase-3 optimistic-concurrency protection is preserved, but I found a few things worth a look:

1. **Unhandled exception path from `build_prompt_context`** — `enqueue_history_compaction`'s `apply_async` call only catches `CeleryError`/`KombuOperationalError`; since it's now invoked unconditionally with no try/except at the call site, an unanticipated exception there can abort an entire agent turn (previously compaction failures were fully isolated).
2. **Lease not refreshed during the actual summarization work** — could let a slow compaction task's lease expire and trigger a redundant duplicate task (wasted LLM cost, not data corruption).
3. **Broad `except Exception` now feeds retries** — `llm_summarise_comms`/`llm_summarise_steps` convert any exception (including non-transient bugs) into a retryable `CompactionSummaryError`, burning retries on errors that will never succeed.
4. **Minor duplication** — the Phase-3 captured-vs-current id race check is duplicated between `compaction.py` and `step_compaction.py`.

Details inline. Nice test coverage on the new lease/coalescing behavior overall.

---

PR #1531, opened 2026-08-10
Add configurable runtime limits for interactive agent processing

matt-greathouse:
## Summary
- Add an interactive agent runtime limit with a 300-second default.
- Propagate explicit runtime limits through task and event-processing layers.
- Schedule runtime-limit follow-ups using the provided delay and queue controls.
- Record runtime termination telemetry and handle budget-cycle cleanup.
- Add coverage for boundary expiry, follow-up behavior, budget preservation, and task propagation.

## Testing
- Added and updated targeted unit tests for runtime-limit handling and follow-up scheduling.


api/agent/core/event_processing.py
```diff
@@ -7348,37 +7356,58 @@ def _refresh_inbound_routing_scope(generation: int) -> bool:
...
                     )
                 except DatabaseError:
                     logger.debug(
                         "Failed to persist runtime limit step for agent %s",
                         agent.id,
                         exc_info=True,
                     )
-                pending_settings = get_pending_drain_settings(settings)
+                delay_seconds = max(0, int(max_iterations_followup_delay_seconds or 0))
+                if max_iterations_followup_delay_seconds is None:
```
matt-greathouse, line 7383: The runtime-limit follow-up delay is missing the minimum floor that the sibling max-iterations path enforces.

Here (runtime limit branch):
```python
delay_seconds = max(0, int(max_iterations_followup_delay_seconds or 0))
if max_iterations_followup_delay_seconds is None:
    delay_seconds = get_pending_drain_settings(settings).pending_drain_delay_seconds
```
vs. the max-iterations branch a few hundred lines below (unchanged by this PR):
```python
delay_seconds = max(0, int(max_iterations_followup_delay_seconds or 0))
if max_iterations_followup_delay_seconds is None:
    pending_delay = get_pending_drain_settings(settings).pending_drain_delay_seconds
    delay_seconds = max(int(MAX_ITERATIONS_FOLLOWUP_DELAY_SECONDS), int(pending_delay))
```
`MAX_ITERATIONS_FOLLOWUP_DELAY_SECONDS` is 60s. For any caller that doesn't pass `max_iterations_followup_delay_seconds` explicitly (i.e. the standard/default processing queue — `AGENT_EVENT_PROCESSING_MAX_RUNTIME_SECONDS` defaults to 1800s there), hitting the runtime limit reschedules the follow-up after only `AGENT_EVENT_PROCESSING_PENDING_DRAIN_DELAY_SECONDS` (default 5s) instead of the 60s cool-down max-iterations gets. An agent whose work legitimately takes longer than the runtime budget every cycle will get rescheduled every ~5s instead of every ≥60s, causing tight reprocessing/thrashing (lock churn, bootstrap overhead, DB/Redis round trips) that the max-iterations path was specifically designed to avoid.

Since this PR is the one parameterizing `delay_seconds` for the runtime-limit branch (previously it was always `pending_settings.pending_drain_delay_seconds` unconditionally), it would be worth applying the same `max(MAX_ITERATIONS_FOLLOWUP_DELAY_SECONDS, pending_delay)` floor here for consistency, unless the fast reschedule is intentional for this specific path.

---

matt-greathouse (commented):
Single-pass review of add-interactive-runtime-limit (Agent tool unavailable this session, so no multi-agent fan-out / subagent verify pass ran). Traced the diff line-by-line, checked all callers/callees of the changed functions, ran the new/modified unit tests (all 42 pass), and compared the new runtime-limit branch against the pre-existing, structurally-parallel max-iterations branch in the same function. One inline finding: the runtime-limit follow-up delay is missing the 60s minimum floor that the max-iterations path enforces. Everything else — parameter threading through process_agent_events → _process_agent_events_locked → _run_agent_loop, the new AGENT_INTERACTIVE_MAX_RUNTIME_SECONDS setting, the budget_exhausted/close_cycle mirroring, and the queue-forwarding for scheduled follow-ups — checked out as correct and consistent with existing patterns.

---

PR #1537, opened 2026-08-11
Preserve compute tools and successful batches after runtime errors

matt-greathouse:
## Summary

- Keep `run_command` and `python_exec` available after non-retryable and terminal tool errors.
- Track tool-result success explicitly when building prompt history.
- Prevent failed compute results from replacing the latest successful source batch or adding terminal guidance.
- Add regression coverage for filtering, batch selection, and success detection.

## Testing

- Added and updated focused unit tests covering event processing, prompt context, link references, and tool-result schemas.
- Not run (not requested).

### DIFF
--- a/api/agent/core/prompt_context.py
+++ b/api/agent/core/prompt_context.py
@@ -644,6 +644,21 @@ def _tool_result_status_is_ok(result: object) -> bool:
     return isinstance(payload, dict) and str(payload.get("status") or "").casefold() == "ok"
 
 
+def _tool_result_completed_successfully(tool_status: object, result: object) -> bool:
+    if str(tool_status or "").strip().casefold() != PersistentAgentToolCall.Status.COMPLETE:
+        return False
+    try:
+        payload = result if isinstance(result, dict) else json.loads(str(result or ""))
+    except (TypeError, ValueError):
+        return True
+    if not isinstance(payload, dict):
+        return True
+    result_status = str(payload.get("status") or "").strip().casefold()
+    if not result_status:
+        return True
+    return result_status in {"ok", "success", "complete", "completed"}
+
+
 def _source_url_from_tool_params(
     agent: PersistentAgent,
     tool_name: str,
@@ -749,6 +764,7 @@ def _build_browser_task_tool_result_record(
         result_text=json.dumps(normalized_payload, ensure_ascii=False),
         result_id=str(task.id),
         source_batch_id=str(task.id),
+        succeeded=task.status == BrowserUseAgentTask.StatusChoices.COMPLETED,
     )
 
 
@@ -772,6 +788,7 @@ def _build_mcp_task_tool_result_record(task: PersistentAgentMCPTask) -> ToolCall
         result_text=json.dumps(payload, ensure_ascii=False),
         result_id=str(task.id),
         source_batch_id=str(task.id),
+        succeeded=task.status == PersistentAgentMCPTask.Status.COMPLETED,
     )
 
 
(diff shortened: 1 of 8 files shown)


api/agent/core/prompt_context.py
```diff
@@ -5261,6 +5282,10 @@ def _get_unified_history_prompt(
                             else None
                         ),
                         source_bearing=source_bearing,
+                        succeeded=_tool_result_completed_successfully(
+                            row.get("status"),
```
matt-greathouse, line 5287: The `.values()` query that backfills missing-parent tool call records (lines 5237-5248) doesn't select `status`, so `row.get("status")` here is always `None`. That makes `_tool_result_completed_successfully(None, result_text)` always return `False` for every backfilled parent record, regardless of whether it actually succeeded.

Concretely: when a tool call's `parent_tool_call_id` points to a step outside the already-fetched `steps` set (common for `spawn_web_task`/MCP task parent chains), the parent gets pulled in here with `succeeded` hardcoded to `False`. That excludes genuinely successful parent source results from `source_records`/`is_current_source_batch` in `tool_results.py`, silently dropping their SOURCE ARRAYS/prose guidance and previews from the prompt even though the tool call actually completed.

The sibling query at line 5131-5144 (for directly-fetched steps) does include `"status"` — this one appears to be missing it by oversight. Suggest adding `"status"` to the `.values(...)` call at line 5240-5247 and passing `row.get("status")` through here.

---

api/agent/core/tool_results.py
```diff
@@ -507,8 +514,9 @@ def prepare_tool_results_for_prompt(
         recency_position = recency_positions.get(record.step_id)
         is_fresh_tool_call = record.step_id in fresh_step_ids
         is_current_source_batch = (
```
matt-greathouse, line 516: Dropping the `record.source_batch_id is None` fallback here changes behavior for legacy/backfilled sibling records that never got an explicit `source_batch_id` (e.g. records built directly without going through `_source_batch_id_for_tool_result`, or ones from before batch ids existed).

Previously: `is_current_source_batch = record.source_batch_id is None or source_batch_id == current_source_batch_id` — any record without an explicit batch id was always treated as current.

Now: a record is only "current" if its `source_batch_id` exactly equals `current_source_batch_id`. But when the most-recent source record also has `source_batch_id is None`, `current_source_batch_id` falls back to `current_source_record.result_id or short_id_map.get(step_id, step_id)` (lines 467-473) — a value unique to that single record. So among a group of legacy siblings that all share `source_batch_id=None`, only the single newest one now satisfies the equality check; the earlier siblings get `is_historical_same_tool_source=True`, which zeroes their `context_hint`/`preview_text` and sets `is_inline=False` (lines 583-586) — hiding their content behind a terse "HISTORICAL SOURCE BATCH" note even though they're from the same turn.

This is exactly the scenario `test_scrape_guidance_groups_visible_legacy_siblings_without_batch_ids` constructs (3 same-tool records, no `source_batch_id` set), but that test only asserts on the combined guidance text/id list — it doesn't check whether each individual sibling's preview is actually visible in the prompt vs. suppressed. Worth adding an assertion there that all three previews (not just the work-set guidance note) survive, or restoring the `is None` fallback if this suppression is unintended.

---

matt-greathouse (commented):
Automated code review (high effort). Found 2 issues, both verified directly against the diffed code — no test-suite run (no Python/Django environment in this session).

1. `prompt_context.py:5286` — backfilled parent tool-call records always get `succeeded=False` because the fallback query omits `status`.
2. `tool_results.py:516` — dropping the `source_batch_id is None` fallback in `is_current_source_batch` causes legacy sibling records without explicit batch ids to be wrongly historicized (only the single newest one stays "current"), which the PR's own new test doesn't catch because it only checks combined guidance text, not per-sibling preview visibility.

See inline comments for details and suggested fixes.

---

PR #1540, opened 2026-08-11
Route inbound SMS and Discord processing through low-latency queues

matt-greathouse:
## Summary

- Treat inbound SMS as interactive processing and prefer the low-latency queue.
- Route Discord debounce and fallback processing through the interactive enqueue helper.
- Support Celery countdowns while preserving interactive queue defaults.
- Reduce the default Discord inbound debounce interval from 15 to 5 seconds.
- Update coverage for SMS routing, Discord fallback/eager behavior, countdowns, and queue preferences.

## Testing

- Added and updated targeted unit tests for SMS and Discord inbound processing and interactive task enqueueing.


api/agent/tasks/process_events.py
```diff
@@ -247,6 +251,8 @@ def process_agent_events_task(
     if queue_latency_seconds is not None:
         span.set_attribute("celery.queue_latency_seconds", queue_latency_seconds)
     if task_queue == AGENT_INTERACTIVE_PROCESSING_QUEUE:
+        if prefer_low_latency is None:
```
matt-greathouse, line 254: Forcing `prefer_low_latency = True` whenever a task lands on the interactive queue silently disables the documented per-iteration adaptive low-latency check for WEB chat sessions.

Before this diff, WEB-channel messages were dispatched with `prefer_low_latency` left unset (`None`), so `event_processing.py`'s `_resolve_low_latency_preference` re-checked `has_deliverable_web_session(agent)` on every loop iteration and dropped low-latency routing once the user closed the tab mid-task. After this change, any task queued on `AGENT_INTERACTIVE_PROCESSING_QUEUE` (which WEB always uses) gets `prefer_low_latency` hard-coded to `True` here, and that fixed value is threaded through as `explicit_prefer_low_latency` for the whole run, so the per-iteration `has_deliverable_web_session` fallback never runs again.

Concretely: a user sends a web-chat message that requires many loop iterations, then closes the browser tab — the run keeps getting routed to low-latency endpoints for the rest of the run instead of falling back once the session ends, contrary to the documented "automatically prefer low latency for active web sessions" behavior.

---

api/agent/comms/message_service.py
```diff
@@ -115,10 +115,15 @@ def dispatch(self) -> None:
             from api.agent.tasks import enqueue_interactive_process_agent_events, process_agent_events_task
 
             if is_interactive:
+                interactive_kwargs = {
+                    "inbound_generation": inbound_generation,
+                    "inbound_message_id": self.message_id,
+                }
+                if self.channel == CommsChannel.SMS:
```
matt-greathouse, line 122: Minor: this SMS-only `prefer_low_latency = True` special case looks redundant now. `is_interactive` already routes SMS onto `AGENT_INTERACTIVE_PROCESSING_QUEUE`, and `process_events.py` now defaults `prefer_low_latency` to `True` for every task on that queue (including WEB, which passes nothing). So this branch achieves nothing beyond what the queue-level default already provides — could probably be dropped and let the task-level default apply uniformly.

---

matt-greathouse (commented):
Automated code review. Found one behavioral issue worth confirming (the interactive-queue default silently disabling the per-iteration web-session low-latency fallback) and one minor redundancy (SMS-specific `prefer_low_latency` flag now duplicating the new queue-level default). Details inline.

---

PR #1541, opened 2026-08-11
Migrate Apollo tooling and guard deprecated Pipedream providers

matt-greathouse:
## Summary

- Add Apollo migration support across tool execution, runtime tracking, and system skills
- Introduce feature-flagged guards for deprecated Pipedream Apollo and Google Sheets providers
- Update event processing and analytics integrations
- Add unit coverage for parallel tool execution and provider guard behavior

## Testing

Not run (not requested)

### DIFF
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -3922,6 +3922,7 @@ def _prepare_tool_batch(
             logger.info("Agent %s preparing tool %d/%d: %s", agent.id, idx, len(tool_calls), tool_name)
 
             preflight_entry: Optional[ToolCatalogEntry] = None
+            preflight_entry_resolved = False
             preflight_params: dict[str, Any] = {}
             try:
                 _, parsed_preflight_params = _parse_tool_call_params(
@@ -3940,6 +3941,7 @@ def _prepare_tool_batch(
             )
             if preflight_blocked_integration:
                 preflight_entry = resolve_tool_entry(agent, tool_name)
+                preflight_entry_resolved = True
                 if preflight_entry is not None:
                     preflight_blocked_integration = match_deprecated_pipedream_integration(
                         tool_name,
@@ -3972,7 +3974,8 @@ def _prepare_tool_batch(
                 and not is_credit_message_only_allowed_tool(tool_name)
             )
             if credit_message_only_restricted and not preflight_blocked_integration:
-                preflight_entry = resolve_tool_entry(agent, tool_name)
+                if not preflight_entry_resolved:
+                    preflight_entry = resolve_tool_entry(agent, tool_name)
                 preflight_blocked_integration = match_deprecated_pipedream_integration(
                     tool_name,
                     preflight_params,
--- a/api/agent/core/event_processing.py
+++ b/api/agent/core/event_processing.py
@@ -87,8 +87,8 @@ from api.evals.credit_policy import is_eval_credit_exempt_context
 from api.evals.execution import get_current_eval_routing_profile
 from api.services.deprecated_provider_guard import (
     filter_deprecated_provider_blocked_tool,
+    get_blocked_deprecated_pipedream_integration,
     is_deprecated_provider_blocked_result,
-    is_pipedream_google_sheets_blocked_call,
 )
 from . import internal_reasoning
 from .daily_limit_mode import (
@@ -2437,7 +2437,7 @@ class _PreparedToolExecution:
     parallel_safe: bool
     parallel_ineligible_reason: Optional[str]
     resolved_entry: Optional[ToolCatalogEntry] = None
-    pipedream_google_sheets_blocked: bool = False
+    deprecated_provider_integration: Optional[str] = None
     display_metadata: Dict[str, Any] = field(default_factory=dict)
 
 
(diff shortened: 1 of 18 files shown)


api/services/deprecated_provider_guard.py
```diff
@@ -131,12 +220,17 @@ def filter_deprecated_provider_blocked_tool(
...
         tool_name
         for tool_name in tool_names
-        if pipedream_app_slug_for_tool_name(tool_name) == GOOGLE_SHEETS_INTEGRATION
+        if _integration_for_tool_name(tool_name, {}) == blocked_integration
     }
     blocked_tool_names = {blocked_tool_name}
     for tool_name in candidates - blocked_tool_names:
         entry = _resolve_tool_entry(agent, tool_name)
-        if entry is not None and _entry_targets_pipedream_google_sheets(entry, {}):
+        detected_integration = (
```
matt-greathouse: Behavior change worth double-checking: previously, `filter_deprecated_provider_blocked_tool`'s candidate loop only blocked a tool when `entry is not None and _entry_targets_pipedream_google_sheets(entry, {})` — an unresolved catalog entry was never treated as a match. Now, when `_resolve_tool_entry` returns `None`, it falls back to `_integration_for_tool_name(tool_name, {})`, which matches purely on the `tool_name` prefix before the first `-` (via `pipedream_app_slug_for_tool_name`).

So a transient catalog-resolution miss (or any non-Pipedream tool that happens to be named like `apollo_io-...`) can now get filtered out of the same-turn roster refresh based on name alone, without ever being confirmed as the actual blocked Pipedream MCP entry. Was this fallback intentional, or should unresolved entries still be excluded from `blocked_tool_names` as before?

---

api/agent/core/event_processing.py
```diff
@@ -3966,7 +3922,30 @@ def _prepare_tool_batch(
...
+            preflight_blocked_integration = match_deprecated_pipedream_integration(
+                tool_name,
+                preflight_params,
+            )
+            if preflight_blocked_integration:
+                preflight_entry = resolve_tool_entry(agent, tool_name)
+                if preflight_entry is not None:
+                    preflight_blocked_integration = match_deprecated_pipedream_integration(
+                        tool_name,
+                        preflight_params,
```
matt-greathouse, line 3948: Minor: when the first preflight pass matches an integration by name, resolves `preflight_entry`, and then the entry-based re-check clears `preflight_blocked_integration` back to `None` (e.g. name looked like `apollo_io-*` but the resolved entry isn't actually that Pipedream integration), the later `if credit_message_only_restricted and not preflight_blocked_integration:` block re-resolves the exact same `tool_name` via `resolve_tool_entry` again — a redundant DB/catalog lookup for the same call. Not a correctness bug, just an avoidable double resolution in that one path.

---

api/agent/system_skills/service.py
```diff
@@ -116,10 +115,8 @@ def prepare_google_sheets_native_handoff(agent: PersistentAgent) -> str:
         usage_count=F("usage_count") + 1,
     )
     if not updated:
-        return GOOGLE_SHEETS_HANDOFF_EXPLICITLY_DISABLED
-    return GOOGLE_SHEETS_HANDOFF_READY
-
-
+        return NATIVE_INTEGRATION_HANDOFF_EXPLICITLY_DISABLED
+    return NATIVE_INTEGRATION_HANDOFF_READY
```
matt-greathouse, line 119: Nit: the blank lines between `prepare_native_integration_handoff` and `_system_skill_keys_for_tool` were dropped in this edit (PEP8 wants 2 blank lines between top-level defs). Worth restoring for lint cleanliness.

---

api/services/deprecated_provider_guard.py
```diff
@@ -1,25 +1,79 @@
...
+            "connect Google Drive, and select the spreadsheets to use. Do not retry Pipedream."
+        ),
+        analytics_event=AnalyticsEvent.PIPEDREAM_GOOGLE_SHEETS_EXECUTION_BLOCKED,
+    ),
+    APOLLO_INTEGRATION: DeprecatedPipedreamIntegration(
+        key=APOLLO_INTEGRATION,
+        display_name="Apollo",
+        pipedream_app_slugs=frozenset({"apollo_io", "apollo_io_oauth"}),
+        native_skill_key=APOLLO_NATIVE_SYSTEM_SKILL_KEY,
+        switch_name=PIPEDREAM_APOLLO_GUARD,
```
matt-greathouse, line 63: This generalizes the deprecated-provider *execution* guard to Apollo, but `filter_guarded_pipedream_google_sheets_tools` in `api/services/pipedream_apps.py` (used by `mcp_manager.py:2516` to build the tool roster) is still Sheets-only and untouched by this PR — there's no Apollo counterpart.

Sheets tools get proactively removed from the roster once the native handoff is ready, so the model stops seeing them; Apollo Pipedream tools will stay visible indefinitely and keep getting offered to the model even after `apollo_native` is ready, so the agent likely keeps calling the deprecated tool and hitting the same `deprecated_provider_blocked` error on repeat turns instead of transitioning once, the way Sheets does.

Is a roster-level filter for Apollo planned as a fast-follow, or should it land with this change?

---

matt-greathouse (commented):
Automated code review via `/code-review`. All findings were manually verified against the diff before posting. Summary:

- **Roster gap**: Apollo Pipedream tools aren't proactively hidden from the tool list once the native handoff is ready, unlike Sheets — likely repeat `deprecated_provider_blocked` errors instead of a one-time transition.
- **Guard fallback**: `filter_deprecated_provider_blocked_tool`'s same-turn refresh now matches unresolved tool entries by name-prefix alone, a behavior change from the previous entry-confirmed-only logic.
- **Minor redundancy**: a preflight code path can re-resolve the same tool entry twice in one call.
- **Nit**: missing blank-line spacing between two top-level functions (PEP8).

Nothing blocking; flagging for author judgment on the roster gap and guard-fallback change in particular.

---

PR #1542, opened 2026-08-11
Add managed OAuth MCP integrations for HubSpot

matt-greathouse:
## Summary

- Add managed MCP integration support with OAuth connection and credential handling
- Register HubSpot managed MCP tools, system skills, metadata, and error normalization
- Add HubSpot insights UI support and native integration API wiring
- Document HubSpot setup and add evals and unit coverage
- Include database migrations for managed MCP integration state

## Testing

- Added and ran focused unit tests for managed MCP integrations and MCP tools
- Added HubSpot managed MCP evaluation scenarios and tests


api/services/pipedream_apps.py
```diff
@@ -62,6 +62,7 @@ class PipedreamOwnerAppsState:
 class PipedreamAppVisibility:
     deprecated_app_slugs: frozenset[str]
     connected_app_slugs: frozenset[str] = frozenset()
+    suppressed_app_slugs: frozenset[str] = frozenset()
```
matt-greathouse, line 66: `suppressed_app_slugs` is only threaded through `get_pipedream_app_visibility_for_agent` and `get_effective_pipedream_app_slugs_for_agent`. `serialize_owner_apps_state` (used by the Apps settings API) still builds `PipedreamAppVisibility(frozenset(get_deprecated_pipedream_app_slugs()))` without it, and `filter_deprecated_pipedream_apps_without_agent` likewise omits it.

That means the Settings > Apps UI will still list/allow selecting HubSpot as a Pipedream app and report it as "selected"/"effective", while tool resolution for agents (which does apply the suppression) silently filters out every `hubspot-*` Pipedream tool. Users would see a connected integration that never actually produces usable tools, with no indication why.

Worth wiring `suppressed_app_slugs=managed_mcp_suppressed_pipedream_app_slugs()` into all `PipedreamAppVisibility` construction sites, not just the agent-facing ones.

---

api/services/managed_mcp_integrations.py
```diff
@@ -0,0 +1,643 @@
...
+    owner_org = agent.organization if agent.organization_id else None
+    return {
+        provider.key
+        for provider in MANAGED_OAUTH_MCP_PROVIDERS.values()
+        if resolve_managed_mcp_connection_mode(provider.key, owner_user, owner_org)
+        == ManagedMCPConnectionMode.MANAGED_MCP
+    }
+
+
+def managed_mcp_suppressed_pipedream_app_slugs() -> set[str]:
```
matt-greathouse: This suppresses `hubspot`'s Pipedream tools for *every* agent platform-wide as soon as this deploys — it iterates all `MANAGED_OAUTH_MCP_PROVIDERS` unconditionally, with no `owner_user`/`owner_org` scoping and no check of `resolve_managed_mcp_connection_mode`/the `hubspot_mcp` waffle rollout for that specific owner.

Compare with `is_app_visible` (pipedream_apps.py), where `suppressed_app_slugs` membership short-circuits to `False` *before* the `connected_app_slugs` grandfather check that the parallel `deprecated_app_slugs` path respects. So a workspace that isn't in the `hubspot_mcp` rollout at all, but has an existing working Pipedream HubSpot connection, would lose access to those tools immediately — with no rollout gate protecting them.

Should this be scoped per-owner (checking `resolve_managed_mcp_connection_mode(...) == MANAGED_MCP` for that provider/owner) rather than suppressing globally based only on provider registration?

---

api/agent/tools/mcp_manager.py
```diff
@@ -1215,9 +1231,13 @@ def _ensure_runtime_oauth(
     ) -> tuple[MCPServerRuntime, Optional[Dict[str, Any]]]:
         if server.auth_method != MCPServerConfig.AuthMethod.OAUTH2:
             return server, None
-        if server.oauth_access_token and (
-            server.oauth_expires_at is None
-            or server.oauth_expires_at > timezone.now() + OAUTH_REFRESH_SAFETY_MARGIN
+        if (
+            not server.metadata.get("managed_oauth")
```
matt-greathouse: By adding `not server.metadata.get("managed_oauth")` to this guard, managed-OAuth servers now always fall through to `ensure_mcp_oauth_credential`, which reloads the config from the DB, re-resolves `resolve_managed_mcp_connection_mode` (another DB query), and takes a Redlock — even when the cached in-memory token still has plenty of validity left. Non-managed OAuth servers keep the fast, no-DB-access path.

For an agent making several HubSpot tool calls back-to-back in one turn, this adds a DB round trip + Redis lock coordination on every single call instead of just when a refresh is actually needed. Was this intentional (e.g. to keep managed credentials strictly in sync), or could the freshness fast-path be preserved for managed servers too when the cached token isn't near expiry?

---

api/services/managed_mcp_integrations.py
```diff
@@ -0,0 +1,643 @@
...
+        "tickets, and other available CRM tools."
+    ),
+    server_url="https://mcp.hubspot.com/",
+    authorization_endpoint="https://mcp.hubspot.com/oauth/authorize/user",
+    token_endpoint="https://mcp.hubspot.com/oauth/v3/token",
+    token_endpoint_auth_method="client_secret_post",
+    scopes=(),
+    icon="hubspot",
+    waffle_flag=HUBSPOT_MCP_WAFFLE_FLAG,
+    system_skill_key="hubspot_native",
```
matt-greathouse: `system_skill_key="hubspot_native"` is a hardcoded string literal that duplicates `HUBSPOT_NATIVE_SYSTEM_SKILL_KEY` from `api.agent.system_skills.defaults` (which is used at `defaults.py:782` for the actual skill registration). If that constant is ever renamed, `trigger_agents_for_managed_mcp_change`'s `PersistentAgentSystemSkillState.objects.filter(skill_key="hubspot_native", ...)` query would silently stop matching any rows — no test failure, agents just stop waking on HubSpot connect/rollout changes. Minor, but worth importing the constant instead of re-typing the literal.

---

console/native_integrations_api.py
```diff
@@ -301,6 +385,53 @@ def post(self, request: HttpRequest, provider_key: str, *args: Any, **kwargs: An
 
         try:
             owner_scope, owner_user, owner_org = _resolve_native_integration_owner(request)
+        except PermissionDenied as exc:
+            return _permission_denied_response(exc)
+
+        managed_session = managed_mcp_oauth_session_exists(provider.key, state, request.user)
+        if managed_mcp_provider_enabled(provider.key, owner_user, owner_org) or managed_session:
```
matt-greathouse: This branches on `managed_mcp_provider_enabled(...)` evaluated *at callback time*, rather than solely on whether a managed session actually exists (`managed_session`). Edge case: a user starts the legacy (non-managed) HubSpot connect flow while `hubspot_mcp` is disabled for their workspace (no `MCPServerOAuthSession` row is created). If the flag gets enabled for that workspace before they complete HubSpot's consent screen, this condition becomes `True` on the `managed_mcp_provider_enabled` disjunct alone, routing into `complete_managed_mcp_oauth` → `_consume_managed_session`, which finds no matching session row and raises `ValidationError` — discarding the authorization code the user just obtained and forcing them to restart the flow.

Might be worth keying this decision off `managed_session` primarily (i.e., trust how the flow was actually initiated) rather than re-evaluating the flag at callback time.

---

matt-greathouse (commented):
Automated code review of the managed OAuth MCP / HubSpot integration changes. All findings below are verified against the diff — line numbers, call sites, and control flow were traced directly in the branch.

**Summary:** 5 findings, roughly in order of impact:
1. `suppressed_app_slugs` isn't wired into all `PipedreamAppVisibility` construction sites (Settings UI can show/select HubSpot as a Pipedream app while tool resolution silently filters it out).
2. `managed_mcp_suppressed_pipedream_app_slugs()` suppresses Pipedream HubSpot tools globally, unscoped by owner or the `hubspot_mcp` rollout flag — could break already-connected users outside the rollout.
3. The managed-OAuth token freshness fast-path is skipped entirely, adding a DB reload + Redis lock coordination to every managed-integration tool call.
4. A hardcoded `"hubspot_native"` string duplicates a constant that already exists — minor maintainability risk.
5. An OAuth callback edge case where toggling the `hubspot_mcp` flag mid-flow can discard an already-obtained authorization code.

No blocking correctness issues in the happy path — these are mostly edge cases and consistency gaps worth a look before merge.

---

PR #1543, opened 2026-08-11
Reduce agent response latency with incremental snapshots and atomic run tracking

matt-greathouse:
## Summary
- Persist agent process run counts and record runs atomically to reduce repeated counting queries.
- Update SQLite message snapshots incrementally, preserving unchanged rows and pruning removed messages.
- Add tracing and slow-lock diagnostics for prompt snapshots and organization credit consumption.
- Add unit coverage for run-count initialization, rollback behavior, and SQLite synchronization.

## Testing
- Added and updated targeted unit tests for event processing and SQLite message snapshots.
- Not run (not requested).


api/agent/core/event_processing.py
```diff
@@ -7044,25 +7069,11 @@ def _process_agent_events_locked(
...
-            processing_step = PersistentAgentStep.objects.create(
-                agent=agent,
-                description="Process events",
-            )
-            sys_step = PersistentAgentSystemStep.objects.create(
-                step=processing_step,
-                code=PersistentAgentSystemStep.Code.PROCESS_EVENTS,
-            )
+        with tracer.start_as_current_span("Agent Bootstrap Run Count"):
+            processing_step, is_first_run, run_sequence_number = _record_process_run_start(agent)
```
matt-greathouse: **Critical: this refactor leaves a dangling `sys_step` reference that will crash every successful agent run.**

The removed code above used to assign `sys_step = PersistentAgentSystemStep.objects.create(...)` directly. `_record_process_run_start` now creates the `PersistentAgentSystemStep` internally (lines 5521-5524) but only returns `processing_step` (a `PersistentAgentStep`) — the system step reference is discarded.

A few lines further down (currently `event_processing.py:7109-7114`, just after `_run_agent_loop(...)` returns), the old code still does:
```python
sys_step.notes = "simplified"
try:
    sys_step.save(update_fields=["notes"])
except OperationalError:
    close_old_connections()
    sys_step.save(update_fields=["notes"])
```
`sys_step` is never assigned anywhere in this scope anymore, so this raises `NameError`. Since it's inside a bare `try/finally` (no `except`), the error propagates to the caller's exception handler, which logs an error, closes the agent's budget cycle as a failure, and re-raises — **turning every successful agent-processing run into a reported failure**, and the `"simplified"` notes marker never gets persisted.

Fix: have `_record_process_run_start` also return the created `PersistentAgentSystemStep` (or expose it via `processing_step.system_step`) and update the call site to use that instead of `sys_step`.

---

matt-greathouse (commented):
Automated code review via `/code-review`.

Found one critical, high-confidence bug (see inline comment): a dangling reference to the removed `sys_step` local variable in `_process_agent_events_locked` (`event_processing.py`) after the `_record_process_run_start` refactor — this will raise `NameError` on every successful agent-processing run.

Everything else reviewed came back clean: the `process_run_count` lazy-init/locking logic (matches its new tests), the SQLite messages upsert/diff rewrite in `message_results.py` (column ordering, rebuild detection, and delete-chunking all check out), the `tasks/services.py` credit-consumption tracing refactor (control flow preserved), and the new migration/model field.

---

PR #1545, opened 2026-08-11
Limit burn-rate judge suggestions to one run per UTC day

matt-greathouse:
## Summary

- Add a per-agent daily cache-backed limit for burn-rate tier step-down judge runs.
- Preserve independent automatic judge triggers when the burn-rate limit is reached.
- Add coverage for daily limiting and UTC-date reset behavior.

## Testing

- Added unit tests covering the daily cap, independent triggers, and date-based reset.

### DIFF
--- a/api/agent/core/agent_judge.py
+++ b/api/agent/core/agent_judge.py
@@ -152,13 +152,8 @@ def maybe_run_agent_judge(
             BURN_RATE_TIER_STEP_DOWN_REASON in trigger.reasons
             and not _claim_burn_rate_judge_daily_slot(agent)
         ):
-            trigger = build_judge_trigger(
-                agent,
-                tools=tools,
-                extra_trigger_reasons=extra_trigger_reasons,
-                trigger_context=trigger_context,
-            )
-            if trigger is None:
+            trigger.reasons.remove(BURN_RATE_TIER_STEP_DOWN_REASON)
+            if not trigger.reasons:
                 return
         _run_judge(agent, trigger, routing_profile=routing_profile)
     except Exception:
@@ -306,7 +301,7 @@ def build_judge_trigger(
         _trigger_reasons(recent_messages, recent_tool_calls),
         extra_trigger_reasons or [],
     )
-    if BURN_RATE_TIER_STEP_DOWN_REASON in reasons and _has_burn_rate_judge_daily_slot(agent):
+    if BURN_RATE_TIER_STEP_DOWN_REASON in reasons and _burn_rate_judge_daily_slot_claimed(agent):
         reasons.remove(BURN_RATE_TIER_STEP_DOWN_REASON)
     if not reasons:
         return None
@@ -1467,7 +1462,7 @@ def _burn_rate_judge_daily_cache_timeout(now=None) -> int:
     return max(1, int((next_day - current_time).total_seconds()))
 
 
-def _has_burn_rate_judge_daily_slot(agent: PersistentAgent) -> bool:
+def _burn_rate_judge_daily_slot_claimed(agent: PersistentAgent) -> bool:
     try:
         return bool(get_redis_client().get(_burn_rate_judge_daily_cache_key(agent)))
     except redis.RedisError:
--- a/api/agent/core/agent_judge.py
+++ b/api/agent/core/agent_judge.py
@@ -49,6 +49,7 @@ JUDGE_RUN_CACHE_TTL_SECONDS = 60 * 60 * 12
 JUDGE_MIN_STEP_GAP = 10
 JUDGE_RUN_COOLDOWN_SECONDS = 60 * 45
 JUDGE_DAILY_RUN_LIMIT = 6
+BURN_RATE_TIER_STEP_DOWN_REASON = "burn_rate_tier_step_down"
 REPORT_TOOL_NAME = "report_judge_suggestion"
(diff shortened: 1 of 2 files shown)


api/agent/core/agent_judge.py
```diff
@@ -1433,6 +1448,34 @@ def _judge_run_cache_key(agent: PersistentAgent, evidence_hash: str) -> str:
...
+
+def _has_burn_rate_judge_daily_slot(agent: PersistentAgent) -> bool:
+    return bool(cache.get(_burn_rate_judge_daily_cache_key(agent)))
+
+
+def _claim_burn_rate_judge_daily_slot(agent: PersistentAgent) -> bool:
+    return cache.add(
+        _burn_rate_judge_daily_cache_key(agent),
+        True,
+        timeout=_burn_rate_judge_daily_cache_timeout(),
```
matt-greathouse: **Daily cap relies on `django.core.cache.cache`, which is not configured as a shared backend in this project.** `config/settings.py` has no `CACHES` setting (Redis is wired up for Channels/Celery only, via `CHANNEL_LAYERS`/`CELERY_BROKER_URL`, not `django.core.cache`), so Django falls back to the process-local `LocMemCache`.

`cache.add`/`cache.get` here will therefore only dedupe within a single worker process's memory. In a production deployment with multiple Celery/gunicorn worker processes, the same agent's tasks can land on different workers over the course of a day, and each worker has its own local cache — so the "once per UTC day" guarantee this PR is meant to add won't actually hold cluster-wide; it degrades to "once per worker process per day," which could still allow several burn-rate judge runs for the same agent on the same day.

Worth confirming whether a shared cache backend (e.g. Redis via `django-redis`) is configured in the deployed settings that aren't in this repo, or whether the daily cap should instead be enforced via a DB-backed check (similar to `_judge_completion_count_today`) to get a real cross-process guarantee.

---

api/agent/core/agent_judge.py
```diff
@@ -1433,6 +1448,34 @@ def _judge_run_cache_key(agent: PersistentAgent, evidence_hash: str) -> str:
...
+        hour=0,
+        minute=0,
+        second=0,
+        microsecond=0,
+    )
+    return max(1, int((next_day - current_time).total_seconds()))
+
+
+def _has_burn_rate_judge_daily_slot(agent: PersistentAgent) -> bool:
+    return bool(cache.get(_burn_rate_judge_daily_cache_key(agent)))
```
matt-greathouse: Naming nit: `_has_burn_rate_judge_daily_slot` reads as "a slot is available," but it actually returns `True` when the slot has already been *used up* for the day (i.e. `cache.get` found the claimed marker). That's the opposite of the intuitive meaning and both call sites (`build_judge_trigger` and `maybe_run_agent_judge`) rely on this inverted semantics to work correctly. Something like `_burn_rate_judge_daily_slot_claimed` (or `_burn_rate_judge_daily_limit_reached`) would make the call sites read correctly and reduce the risk of a future edit flipping the condition by mistake.

---

api/agent/core/agent_judge.py
```diff
@@ -144,6 +145,18 @@ def maybe_run_agent_judge(
         )
         if trigger is None:
             return
+        if (
+            BURN_RATE_TIER_STEP_DOWN_REASON in trigger.reasons
+            and not _claim_burn_rate_judge_daily_slot(agent)
```
matt-greathouse, line 153: Minor: on a lost race for the daily slot, `build_judge_trigger` is called a second time in full, re-running `_count_non_judge_steps`, `_latest_judge_completion_created_at`/`_latest_judge_suggestion_created_at`, `_recent_messages`, `_recent_tool_calls`, etc. — a handful of extra DB queries just to drop the one reason. Since this only needs to strip `BURN_RATE_TIER_STEP_DOWN_REASON` from the already-built `trigger.reasons` (and bail if that empties the list), it could avoid the full rebuild, e.g.:
```python
if BURN_RATE_TIER_STEP_DOWN_REASON in trigger.reasons and not _claim_burn_rate_judge_daily_slot(agent):
    trigger.reasons.remove(BURN_RATE_TIER_STEP_DOWN_REASON)
    if not trigger.reasons:
        return
```
Not a correctness issue (this path only fires on a rare claim race), just avoids redundant queries.

---

matt-greathouse (commented):
Reviewed the daily burn-rate judge cap. Logic for merging/removing the trigger reason and the race-safe `cache.add` claim is sound, and the new tests cover the daily cap, the independent-trigger preservation, and the UTC-date reset well.

One finding worth resolving before merge: the daily cap is built on `django.core.cache.cache`, but this repo's `config/settings.py` has no `CACHES` setting (Redis is wired up for Channels/Celery, not Django's cache framework), so it defaults to the process-local `LocMemCache`. That means the "once per UTC day" guarantee only holds within a single worker process, not cluster-wide — see inline comment for details.

Also left a naming nit (`_has_burn_rate_judge_daily_slot`'s return value is inverted from what the name implies) and an optional efficiency suggestion (avoid a full trigger rebuild on the rare claim-race path). Neither blocks merge.

CI note: the one failing shard (`tests (shard_02)`) is unrelated to this change — it's an MCP server version-string assertion (`2.30.1` vs `2.31.0`) in `tests/unit/test_remote_mcp.py`.

---

PR #1547, opened 2026-08-12
Update Debug Mode

matt-greathouse:


frontend/src/components/agentChat/tooling/toolRegistry.ts
```diff
@@ -409,6 +406,9 @@ function buildToolEntries(clusterCursor: string, entry: ToolCallEntry, options:
       sourceEntry: entry,
     }]
   }
+  if (normalizedName === 'send_discord_message' || normalizedName === 'discord_send_message') {
```
matt-greathouse, line 409: This explicit `send_discord_message` / `discord_send_message` check is now dead code. `CHAT_SKIP_TOOL_NAMES` (checked right below at line 412) is built from `COMMUNICATION_TOOL_NAMES`, which already includes both `'send_discord_message'` and `'discord_send_message'` (`frontend/src/components/tooling/toolMetadata.ts:43-49,55-56`). Since this block now sits after the `entry.developerEvent` branch, any normalized name that matches it will also match `CHAT_SKIP_TOOL_NAMES.has(...)` on the next check — this block can be deleted without changing behavior.

---

frontend/src/components/agentChat/MessageEventCard.tsx
```diff
@@ -179,6 +184,7 @@ export const MessageEventCard = memo(function MessageEventCard({
 
   const liveRelativeLabel = useRelativeTimestamp(message.timestamp)
   const relativeLabel = liveRelativeLabel || message.relativeTimestamp || ''
+  const absoluteLabel = formatAbsoluteTimestamp(message.timestamp, timeZone)
```
matt-greathouse, line 187: `absoluteLabel` is computed unconditionally on every render via `formatAbsoluteTimestamp` (which constructs an `Intl.DateTimeFormat`), even though it's only used when `exactTimestamp` is true or `exactTimestampAlongsideStatus` applies. Outside Debug Mode (the common case, `exactTimestamp = false` by default), this is wasted work on the message-list render path that scales with message count. Consider gating it behind `exactTimestamp` (e.g. `const absoluteLabel = exactTimestamp ? formatAbsoluteTimestamp(...) : null`). Minor, but worth a quick fix since it's on a hot path.

---

matt-greathouse (commented):
Automated code review (medium effort) on `update-dev-mode` vs `main`.

Overall this is a clean, well-scoped rename/enhancement of Developer Mode → Debug Mode, with a new exact-timestamp display path and a message-search shortcut refactor. Two small, concrete issues found (inline below): one bit of now-dead code and one minor render-path inefficiency. Nothing blocking.

One additional note not tied to a specific line: this PR adds/expands several front-end test files (`ChatSidebar.test.ts`, `developerTimelineDisplay.test.ts`, `time.test.ts`, plus additions to existing `*.test.tsx` files) even though `AGENTS.md` states "Do not add front-end tests unless specifically requested." Flagging for awareness in case that policy should apply here — the tests themselves look reasonable and well-targeted at the new behavior.

---

PR #1551, opened 2026-08-12
Keep active work running despite future schedules

matt-greathouse:
## Summary

- remove iteration progress and low-iteration warnings from agent prompts so the runtime owns processing boundaries
- clarify that configuring a future schedule does not complete separate, executable work in the active request
- add a real-harness regression eval for continuing active work after an hourly fallback schedule is configured
- add prompt/scorer coverage and make the synthetic batch fixture advance through explicit cursors

## Root cause

The agent saw its iteration counter and treated proximity to the processing limit as a reason to voluntarily sleep, even though the runtime automatically schedules continuation at that boundary. It could also treat creating an hourly schedule as a substitute for work that was still executable in the current request.

## Impact

Agents should continue bounded active work until the runtime pauses them, while still sleeping normally when scheduled or one-off work is genuinely complete. Schedule creation and timing behavior remain intact.

## Validation

- active-work regression before fix: failed 3/3
- active-work regression after final prompt cleanup: passed 3/3 (9/9 tasks)
- focused unit tests: 108 passed
- finished/idle counter-evals: 17/17 tasks passed
- primary recurring-schedule creation: 3/3 runs passed
- multiple recurring schedules: 3/3 runs passed
- durable rule plus multiple schedules: 3/3 reruns passed
- full agent scheduling suite: 31/33 tasks passed; direct creation, exact timer, update, cancel, listing, and guardrail cases passed. The two misses were an ancillary charter persistence case (passed on rerun) and an optional weekly-automation offer that does not create a schedule.
- `git diff --check`

### DIFF
--- a/api/agent/core/prompt_context.py
+++ b/api/agent/core/prompt_context.py
@@ -4219,11 +4219,9 @@ def _get_system_instruction(
     )
     stop_continue_examples = (
         "## Stop/continue\n\n"
-        "Set will_continue_work=true only while this active request has unsent results, unverified constraints, needed tool "
-        "results, immediately executable work, or its own plan cleanup. Completing delivery or config does not finish a "
-        "separate actionable clause in the same request. Set false only when the current request is complete or blocked. "
-        "Future schedules, queued conversations, and their plan items do not by themselves keep a completed turn open; "
-        "they do not justify deferring executable current work.\n"
+        "Set will_continue_work=true while this request has unsent/unverified results, needed tool results, executable "
+        "work, or plan cleanup. Delivery/config does not finish other work. Set false only when complete or blocked; "
+        "schedules neither complete nor defer current work.\n"
         f"{text_only_guidance}"
         "Before final delivery, mark its delivery step done and leave unrelated steps in todo. Never send a "
         "complete answer with true for cleanup: update_plan "
--- a/api/agent/core/prompt_context.py
+++ b/api/agent/core/prompt_context.py
@@ -3143,16 +3143,6 @@ def add_budget_awareness_sections(
             return f"{seconds // 3600}h"
         return f"{seconds // 86400}d"
 
-    if max_iterations and max_iterations > 0:
-        iteration_text = (
-            f"Iteration progress: {current_iteration}/{max_iterations} in this processing cycle."
-        )
-    else:
-        iteration_text = (
-            f"Iteration progress: {current_iteration} with no maximum iterations specified for this cycle."
-        )
-    sections.append(("iteration_progress", iteration_text, 3, True))
-
     try:
         ctx = get_budget_context()
         if ctx is not None:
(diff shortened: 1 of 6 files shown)


api/agent/core/prompt_context.py
```diff
@@ -3441,21 +3431,6 @@ def _format_cost(value: Decimal | Any) -> str:
     except Exception:
         logger.debug("Failed to append tool cost overview to budget awareness.", exc_info=True)
 
-    if max_iterations and max_iterations > 0:
-        try:
-            if (current_iteration / max_iterations) > 0.8:
-                sections.append(
-                    (
-                        "iteration_warning",
```
matt-greathouse, line 3449: The `>80% iterations used → never false-complete, carry unfinished scope forward` warning is removed here with no replacement. This was a targeted guard against false-completion driven by the loop's own iteration budget (agent close to `max_iterations` with real work left). The new stop/continue wording added below only addresses "future schedules aren't an excuse to stop" — a different failure mode. Was the removal intentional, or should an iteration-budget-specific nudge be kept in some form?

---

api/agent/core/prompt_context.py
```diff
@@ -3143,16 +3143,6 @@ def _format_age(delta: timedelta) -> str:
             return f"{seconds // 3600}h"
         return f"{seconds // 86400}d"
 
-    if max_iterations and max_iterations > 0:
-        iteration_text = (
-            f"Iteration progress: {current_iteration}/{max_iterations} in this processing cycle."
```
matt-greathouse, line 3148: Minor cleanup: with both the `iteration_progress` text (here) and the `iteration_warning` block (further down) removed, `current_iteration` and `max_iterations` are now dead parameters of `add_budget_awareness_sections` — nothing in the function body reads them anymore, even though the sole caller in `event_processing.py` still computes and threads them through. Worth dropping the params (or documenting why they're kept) so a future reader doesn't assume they still drive prompt content.

---

matt-greathouse (commented):
Automated review of the iteration-budget prompt changes and the new `effort_active_work_ignores_future_schedule` eval scenario. Two inline notes below on the removed iteration-budget guard and now-dead parameters.

**Likely CI breakage (not inline-commentable — file isn't touched by this diff):** `tests/unit/test_prompt_context_sqlite_guidance.py:337`, `test_low_iteration_warning_keeps_unfinished_work_active`, does `collector.sections["iteration_warning"]` after calling `add_budget_awareness_sections`. This PR deletes the whole `if max_iterations and max_iterations > 0: ... sections.append(("iteration_warning", ...))` block (prompt_context.py, old line ~3444), so that section is never appended and the test should now raise `KeyError: 'iteration_warning'`. This test needs to be updated or removed as part of this change.

Everything else (the new eval scenario, its unit test, and the stop/continue prompt wording) looks solid.

---

api/agent/core/prompt_context.py
matt-greathouse: The removal is intentional. The production trajectory showed the iteration counter/warning itself encouraging premature sleep near the loop limit, while the runtime already carries active work into continuation. This branch removes both iteration cues and validates the desired behavior with a real-harness active-work regression.

PR #1554, opened 2026-08-13
Refine agent continuation and daily reporting behavior

matt-greathouse:
## Summary

- Clarify that future schedules do not end executable work or justify sleeping.
- Separate recurring work cadence from communication/reporting cadence.
- Prevent invented exceptions to strict daily-only reporting preferences.
- Add email and sleep tool guidance updates.
- Add real-harness effort calibration coverage for daily reporting and continued work.

## Testing

- Added unit coverage for reporting boundaries, schedule persistence, and continuation behavior.
- Not run (not requested).


api/evals/scenarios/effort_calibration.py
```diff
@@ -2588,6 +2591,300 @@ def batch_result(next_offset: int) -> dict:
...
+        self.record_task_result(
+            run_id,
+            None,
+            EvalRunTask.Status.PASSED,
+            task_name="inject_prompt",
+            observed_summary="Daily-only continuous-work request completed through the real agent harness.",
+            artifacts={"message": inbound},
+        )
+
+        calls = _tool_calls_for_run(run_id, after=inbound.timestamp)
```
matt-greathouse, line 2809: `calls` is fetched once right after the first (inject_prompt) wake and never refreshed after the cron-triggered second wake, so `verify_active_work_continued` (and the `work_calls` it derives) silently ignores every tool call made during `trigger_off_cadence_work` — the exact wake this check is supposed to validate.

Concretely: if the agent correctly acks in wake 1 (few/no batches) and then correctly continues real work in wake 2 after the hourly cron fires (2+ batches), `active_work_continued(calls, minimum_batches=2, ...)` still only sees wake-1's `calls` and fails the task despite correct behavior. Conversely, a misbehaving wake-2 terminal message (`will_continue_work=False`) is invisible to this stale list too, so the scenario can pass despite the regression it exists to catch.

Compare with the pre-existing pattern at line ~1580 (`_tool_calls_for_run(run_id, after=cron_step.created_at, ...)`), which re-queries after the cron trigger — this scenario should do the same before calling `verify_active_work_continued`.

---

api/evals/scenarios/effort_calibration.py
```diff
@@ -2588,6 +2591,300 @@ def batch_result(next_offset: int) -> dict:
...
+                            ),
+                            "verification": (
+                                "This receipt is authoritative; more bounded batches are immediately available."
+                            ),
+                            "notable_finding": (
+                                "Material correction: a prior working assumption is contradicted by the verified "
+                                "batch, but the correction can be included in the next scheduled owner report."
+                            ),
+                        },
+                    }
```
matt-greathouse, line 2767: Minor: `strict_daily_config_respected`'s `has_hourly_work` check reimplements the same recurring-hourly-schedule detection that `EffortActiveWorkIgnoresFutureScheduleScenario.run()` already has inline (~line 2551-2558), but via a different mechanism (set-intersection over `additional_schedules`/`agent.schedule` vs. a `.filter(...).exists()` query). Not a bug, but the two checks can drift out of sync over time (e.g. one gains support for `@hourly` and the other doesn't), silently changing pass/fail behavior for only one scenario. Consider factoring into a shared helper, e.g. `_has_recurring_expression(agent, {"0 * * * *", "@every 1h"})`.

---

matt-greathouse (commented):
Automated code review via `/code-review`. One real bug found: a stale query result in the new eval scenario undermines the assertion it's meant to validate (see inline comment). One minor duplication note also inline. Everything else in the diff (prompt-context prose changes, tool-description tweaks, and matching test assertions) checked out clean.

---

PR #1556, opened 2026-08-17
Agent Exporter

matt-greathouse:
Quitting the imports and focusing just on the exports.


api/services/portable_agent_export_archive.py
```diff
@@ -0,0 +1,951 @@
...
+                schema_rows = connection.execute(
+                    "SELECT type, name, tbl_name, sql FROM sqlite_master WHERE sql IS NOT NULL ORDER BY type, name"
+                ).fetchall()
+                tables = []
+                for (table_name,) in connection.execute(
+                    "SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' ORDER BY name"
+                ).fetchall():
+                    columns = [
+                        {"name": row[1], "type": row[2], "notNull": bool(row[3]), "default": row[4], "primaryKey": bool(row[5])}
+                        for row in connection.execute(f"PRAGMA table_info({json.dumps(table_name)})").fetchall()
```
matt-greathouse: SQLite table names are interpolated into `PRAGMA table_info(...)` using `json.dumps()` instead of proper SQL identifier escaping. `json.dumps` backslash-escapes embedded quotes, but SQLite identifiers require *doubling* embedded double quotes — so a table name containing a `"` (e.g. from `CREATE TABLE "a""b" (...)`) produces invalid SQL and raises `sqlite3.OperationalError`. The row-count query two lines above (line 756) escapes the identifier correctly, which suggests this is an inconsistency/oversight rather than intentional. The surrounding `except` (line 763) then silently drops the whole SQLite export section for that agent.

---

api/agent/tools/sqlite_state.py
```diff
@@ -1285,6 +1285,14 @@ def agent_sqlite_db(agent_uuid: str):  # noqa: D401 – simple generator context
         yield db_path
 
 
+def write_agent_sqlite_export_snapshot(agent_uuid: str, destination_path: str) -> None:
```
matt-greathouse, line 1288: `write_agent_sqlite_export_snapshot` reuses the mutating `agent_sqlite_db()` context manager for what should be a read-only export. That means every export item unconditionally goes through the full restore → VACUUM → zstd-compress → re-upload cycle in `agent_sqlite_db`'s `finally` block, and it contends for the same per-agent Redlock used by live tool execution.

For a bulk export covering agents that are mid-task, this can either block the agent's live SQLite tool call until the export's persistence cycle finishes, or make the export item fail with `AgentSQLiteBusy` if the agent already holds the lock — turning a supposed read-only backup into a source of write amplification and lock contention on production agents. Worth confirming this is intentional (e.g. to guarantee a consistent flushed snapshot) vs. an accidental reuse of the write path where a read-only connection/snapshot would suffice.

---

api/tasks/portable_agent_exports.py
```diff
@@ -0,0 +1,536 @@
...
+        storage = portable_agent_export_storage()
+        with archive_path.open("rb") as archive:
+            saved_key = storage.save(storage_key, File(archive))
+
+        export.refresh_from_db()
+        if not user_can_access_export(export.requester, export):
+            export.storage_key = saved_key
+            revoke_export_artifact(export)
+            return
+        warning_count = sum(item.warning_count for item in items) + len(items) - len(successful)
```
matt-greathouse: `warning_count` folds failed-agent count into the same number surfaced as "warnings": `sum(item.warning_count) + len(items) - len(successful)`. So a failed agent with zero real warnings still adds 1 to `warning_count`.

Example: 2 agents, one succeeds with 0 warnings, one fails outright → `warning_count = 0 + 2 - 1 = 1`. `PortableAgentExportSection.tsx` then renders something like "1 warning · 1 agent could not be exported," implying a distinct redaction/file warning happened in addition to the failure, when there were actually zero warnings. Consider tracking failed-agent count separately from warning_count so the UI doesn't double-represent one failure as both a warning and a failure.

---

api/services/portable_agent_export_archive.py
```diff
@@ -0,0 +1,951 @@
...
+                    ]
+                    row_count = connection.execute(f'SELECT COUNT(*) FROM "{table_name.replace(chr(34), chr(34) * 2)}"').fetchone()[0]
+                    tables.append({"name": table_name, "columns": columns, "rowCount": row_count})
+            finally:
+                connection.close()
+            schema_sql = "\n\n".join(row[3].rstrip(";") + ";" for row in schema_rows) + "\n"
+            (state_dir / "schema.sql").write_text(schema_sql, encoding="utf-8")
+            write_json_file(state_dir / "tables.json", {"tables": tables})
+        except (OSError, sqlite3.Error, SQLiteStateError, AgentSQLiteBusy) as exc:
```
matt-greathouse: This `except` clause (`OSError, sqlite3.Error, SQLiteStateError, AgentSQLiteBusy`) doesn't cover lock-backend connectivity errors (e.g. redis/pottery connection errors) that the Redlock in `agent_sqlite_execution` can raise while acquiring the per-agent lock. The same gap exists in the outer `ARCHIVE_BUILD_ERRORS` tuple in `api/tasks/portable_agent_exports.py` (lines 55-65).

If Redis is briefly unreachable during `write_agent_sqlite_export_snapshot`'s lock acquisition, the resulting exception would propagate uncaught out of `_process_portable_agent_export`, leaving the export stuck in `RUNNING` instead of being marked `FAILED` and degrading gracefully like other storage failures. Worth confirming whether lock-backend errors are already covered elsewhere, or adding them to the caught exception set.

---

matt-greathouse (commented):
Automated review of the portable agent export feature. Found 4 issues worth a look, all in `api/services/portable_agent_export_archive.py`, `api/agent/tools/sqlite_state.py`, and `api/tasks/portable_agent_exports.py`:

1. Incorrect SQL identifier escaping for SQLite table names in schema introspection (uses `json.dumps` instead of doubling embedded quotes).
2. Export snapshotting reuses the mutating/locking `agent_sqlite_db()` path, causing unnecessary write amplification and lock contention with live agent tool execution.
3. `warning_count` conflates failed-agent counts with actual warnings, causing the UI to double-represent a single failure.
4. Exception handling for SQLite export doesn't cover lock-backend (Redis) connectivity errors, which could leave an export stuck in `RUNNING` state.

See inline comments for details and specific failure scenarios.

---

